PR #1, opened 2022-09-01
radicle-git init

FintanH:
I am porting over all the git portions of radicle-link. Currently, a work in progress as I'm running into some strange test failures. Also some other meta setup of the repository.

PR #3, opened 2022-09-08
copying radicle-surf into radicle-git

keepsimple1:
This patch is to copy the current [radicle-surf](https://github.com/radicle-dev/radicle-surf) repo (commit [b85d2183d786e5fa447aab9d2f420a32f1061bfa](https://github.com/radicle-dev/radicle-surf/commit/b85d2183d786e5fa447aab9d2f420a32f1061bfa)) into `radicle-git` repo, as a starting point of migrating surf into this repo and making changes later. 

I made a few small changes as needed to pass the local ci run test. I will add brief comments for these changes in the diff. Otherwise, no other code changes.

For the GitHub workflow, I only set up a minimal one to run the ci. We can improve that in following / other PRs.

~~The whole code base is under `surf/` subdirectory so that they don't interfere with other migration effort like #1 at this point.~~

I have opted for a flat hierarchy for the code organization: 

- both `surf` and `source` are in the root directory workspace. 
- named directory name to be exactly same as the crate names, i.e. `radicle-surf` and `radicle-source`.
- not including `nix` related files as I have not tested that locally and consider `nix` a more generic component for the whole repo.
- `cargo build`, `cargo test`, and `ci/run` works.


surf/surf/src/vcs/git.rs
```diff
@@ -1149,8 +1149,8 @@ mod tests {
     #[test]
     // An issue with submodules, see: https://github.com/radicle-dev/radicle-surf/issues/54
     fn test_submodule_failure() {
-        let repo = Repository::new("..").unwrap();
-        let browser = Browser::new(&repo, Branch::local("master")).unwrap();
+        let repo = Repository::new("../..").unwrap();
```
keepsimple1, line 1152: An enhancement to consider:  `Repository::new` or a different method to use `git2::Repository::open_ext` so that it can automatically search up the parent directories.

surf/surf/src/vcs/git.rs
```diff
@@ -1149,8 +1149,8 @@ mod tests {
...
+        let browser = Browser::new(&repo, Branch::local("main")).unwrap();
```
keepsimple1, line 1153: Consider a change: how about `Browser::new()` to use the default `master` or `main` branch automatically? or `Branch` provide a method `Branch::default()` ?

surf/surf/Cargo.toml
```diff
@@ -37,7 +37,7 @@ default-features = false
 features = []
 
 [dev-dependencies]
-pretty_assertions = "0.6"
+pretty_assertions = "1.3.0"
```
keepsimple1, line 40: This is to address an advisory warning.

surf/deny.toml
```diff
@@ -47,6 +47,7 @@ allow = [
     "GPL-2.0",
     "GPL-3.0",
     "MIT",
+    "Unicode-DFS-2016",
```
keepsimple1, line 50: This is to fix a `cargo deny` license error.

deny.toml
```diff
@@ -0,0 +1,215 @@
...
+# See https://spdx.org/licenses/ for list of possible licenses
+# [possible values: any SPDX 3.7 short identifier (+ optional exception)].
+allow = [
+    "Apache-2.0",
+    "BSD-2-Clause",
+    "CC0-1.0",
+    "GPL-2.0",
+    "GPL-3.0",
+    "MIT",
+    "Unicode-DFS-2016",
```
keepsimple1, line 50: This is to fix a cargo deny license error.

radicle-surf/Cargo.toml
```diff
@@ -0,0 +1,52 @@
...
+serde = { features = ["serde_derive"], optional = true, version = "1" }
+thiserror = "1.0"
+
+[dependencies.git2]
+version = ">= 0.12"
+default-features = false
+features = []
+
+[dev-dependencies]
+pretty_assertions = "1.3.0"
```
keepsimple1, line 40: This is to address an advisory warning.

radicle-surf/build.rs
```diff
@@ -34,7 +34,7 @@ enum Command {
 impl Command {
     fn new() -> io::Result<Self> {
         let current = env::current_dir()?;
-        Ok(if current.ends_with("surf") {
+        Ok(if current.ends_with("radicle-surf") {
```
keepsimple1, line 37: This is needed as we are naming the directory the exact same name as the crate name.

surf/surf/src/vcs/git.rs
FintanH: Ya, that would be a great shout! I would probably call it something like `Repository::find` :)

surf/surf/src/vcs/git.rs
FintanH: I don't think, in general, that there is a "default" branch when it comes to git repositories, or perhaps I'm mistaken? I don't think a library needs to be opinionated on what string is the canonical branch name.

FintanH:
> For the GitHub workflow, I only set up a minimal one to run the ci. We can improve that in following / other PRs.

Magic, how'd you get it to be picked up? :sweat_smile: I can't tell why #1 doesn't have tests running.

> ~The whole code base is under `surf/` subdirectory so that they don't interfere with other migration effort like #1 at this point.~
> 
> I have opted for a flat hierarchy for the code organization:
> 
>     * both `surf` and `source` are in the root directory workspace.
> 
>     * named directory name to be exactly same as the crate names, i.e. `radicle-surf` and `radicle-source`.
> 
>     * not including `nix` related files as I have not tested that locally and consider `nix` a more generic component for the whole repo.
> 
>     * `cargo build`, `cargo test`, and `ci/run` works.

I think I would prefer a separate workspace because of #1. What to call it is another ~bikeshed~ question though :upside_down_face: Here's why:
* I think we'll generally want to consolidate the `LICENSE`, `DCO`, and other top-level meta files, while the other files are specific to `surf` and `source`, e.g. `DEVELOPMENT.md`
* The testing is slightly different, so I think it would be easier to separate them at first and then figure out if/how we want to consolidate them
* These two crates are separate conceptually, in my mind. They're a higher-level API to provide some UI functionality, whereas the others are still lower-level in comparison.

That's my 2c, wdyt?

surf/surf/src/vcs/git.rs
cloudhead: Well, `master` is the default branch in git, and then there's also the `init.defaultBranch` config to change branch is used when you create a new repo. I don't know if this is helpful here but :man_shrugging:

cloudhead:
One thing for the long run: `source` and `surf` should probably get merged, I think they are separate only due to circumstance, and not by design.

surf/surf/src/vcs/git.rs
keepsimple1: That's true. Git would give hints regarding the default branch during init: 

```
$ git init git-test
hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint:
hint: 	git config --global init.defaultBranch <name>
hint:
hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
hint: 'development'. The just-created branch can be renamed via this command:
hint:
hint: 	git branch -m <name>
Initialized empty Git repository in /Users/hanxu/work/test/git-test/.git/
```

keepsimple1:
> Magic, how'd you get it to be picked up? 😅 I can't tell why #1 doesn't have tests running.

Is it because of this error? https://github.com/radicle-dev/radicle-git/actions/runs/3014568317

[Invalid workflow file: .github/workflows/ci.yaml#L101](https://github.com/radicle-dev/radicle-git/actions/runs/3014568317/workflow)

The workflow is not valid. .github/workflows/ci.yaml (Line: 101, Col: 12): Job 'test-macos' depends on unknown job 'build-macos'.

> * I think we'll generally want to consolidate the `LICENSE`, `DCO`, and other top-level meta files, while the other files are specific to `surf` and `source`, e.g. `DEVELOPMENT.md`

Agreed. And this does not conflict with a flat workspace.

> * The testing is slightly different, so I think it would be easier to separate them at first and then figure out if/how we want to consolidate them

I actually like to focus more on `tests/` under each crate. (I.e. `radicle-surf/tests`,  `radicle-source/tests`, assuming we are not merging them yet).

If we have a `test/` under the root directory, I think it's good to cover all crates in this repo. 

> * These two crates are separate conceptually, in my mind. They're a higher-level API to provide some UI functionality, whereas the others are still lower-level in comparison.

Still I don't think such difference would conflict with a flat workspace. Btw, here is a blog I liked related to this topic: https://matklad.github.io/2021/08/22/large-rust-workspaces.html , which talks about the benefits of a flat layout.

surf/surf/src/vcs/git.rs
FintanH: Ya, so if you want to add a function that looks at `init.defaultBranch` I think that would be fine. But `new` should stay as-is imo

FintanH:
> Is it because of this error? https://github.com/radicle-dev/radicle-git/actions/runs/3014568317

Amazing! I didn't spot that :)

> If we have a `test/` under the root directory, I think it's good to cover all crates in this repo.

So once #1 is merged, it will be the single `test` crate and then `surf` would have to organise its tests under `surf/t`, which is fine, but just needs to be considered.

> Still I don't think such difference would conflict with a flat workspace. Btw, here is a blog I liked related to this topic: https://matklad.github.io/2021/08/22/large-rust-workspaces.html , which talks about the benefits of a flat layout.

So the workspace will still be flat, but in the top-level `Cargo.toml` we would have:
```toml
[workspace]
members = [
  "git-ext",
  "git-ref-format",
  "git-trailers",
  "git-types",
  "link-git",
  "macros",
  # TODO: port gitd-lib over
  # "cli/gitd-lib",
  "std-ext",
  "test",
  
  # N.B. Naming for comedy purposes
  "watersports/surf",
  "watersports/source",
]
```

We just have a perceived hierarchy, because we're adding a directory.

keepsimple1:
> One thing for the long run: `source` and `surf` should probably get merged, I think they are separate only due to circumstance, and not by design.

+1.  We plan to do that in its own PR to have a clearer diff.

keepsimple1:
I've rebased the diff to the latest main branch.

FintanH (dismissed):
Sweet, let's merge and start phase 2 :)

FintanH:
> Sweet, let's merge and start phase 2 :)

I spoke too soon, you're missing the DCO sign-off :sweat_smile: It'd also be good to practice to start signing with an [SSH][0] or [GPG][1] key associated with your account too.

I have these lines in my `.gitconfig`:
```
[commit]
	gpgsign = true # signs with gpg when I commit
[alias]
	logs = log --show-signature # shows signatures when I run `git logs`
	commits = commit -s -v # adds my sign-off and shows diff when I run `git commits`
```

[0]: https://calebhearth.com/sign-git-with-ssh
[1]: https://docs.github.com/en/authentication/managing-commit-signature-verification/telling-git-about-your-signing-key

keepsimple1:
> I spoke too soon, you're missing the DCO sign-off 😅 It'd also be good to practice to start signing with an [SSH](https://calebhearth.com/sign-git-with-ssh) or [GPG](https://docs.github.com/en/authentication/managing-commit-signature-verification/telling-git-about-your-signing-key) key associated with your account too.

Sorry I'm still in the process of establishing this habit. I signed off using SSH keys and pushed the commit again.  Let me know if it's done correctly.

FintanH (approved):

Issue #7, opened 2022-09-19
surf: merge source into surf


FintanH:
Merge the `radicle-source` crate into `radicle-surf`, consolidating any duplicate code.

keepsimple1:
done.

Issue #6, opened 2022-09-19
surf: test structure compliance


FintanH:
The test structure should be that all tests are gathered under a `t` sub-crate, followed by the `test` crate depending on this new crate.

- [x] Create `radicle-surf/t` crate
- [x] Add dependency to `test`
- [x] Organise surf tests under the new crate
- [x] Disable testing in `radicle-surf/Cargo.toml`
- [x] Optional: figure out if we should keep the doc tests as-is or what's the best thing to do with them. Perhaps adding `no_run` and keeping them as useful examples

keepsimple1:
done

Issue #5, opened 2022-09-19
surf: convert git2::Oid to git_ext::Oid


FintanH:
The `git_ext::Oid` type has better trait implementations and should be favoured instead of `git2::Oid`.

- [x] Replace `git2::Oid` with `git_ext::Oid` in `surf`/`source`

FintanH:
Closed via #18

PR #10, opened 2022-09-20
merge radicle-source into radicle-surf

keepsimple1:
This is to fix issue #7

- `branch.rs` , `tag.rs` is duplicated with `radicle-surf/src/vcs/git`, hence removed.

UPDATE: `oid.rs` and `error.rs` also removed after addressing comments.

-  `radicle-source` did not have tests. We will add tests when resolving issue #6 .

Also fixing a CI issue on main branch.


radicle-source/src/branch.rs
```diff
@@ -1,108 +0,0 @@
...
-    branches: Vec<Branch>,
-}
-
-/// Given a path to a repo returns the list of branches and if it is managed by
-/// coco.
-///
-/// # Errors
-///
-/// Will return [`Error`] if the repository doesn't exist.
-pub fn local_state(repo_path: &str, default_branch: &str) -> Result<LocalState, Error> {
```
keepsimple1, line 79: I cannot find any usage of this function in radicle http-api. Hence not keeping it.

radicle-surf/src/commit.rs
```diff
@@ -118,12 +126,12 @@ impl Serialize for Header {
...
 
 /// A selection of commit headers and their statistics.
-#[derive(Serialize)]
+#[cfg_attr(feature = "serialize", derive(Serialize))]
 pub struct Commits {
     /// The commit headers
     pub headers: Vec<Header>,
     /// The statistics for the commit headers
-    pub stats: radicle_surf::vcs::git::Stats,
+    pub stats: crate::vcs::git::Stats,
```
FintanH, line 134: nit: I'd prefer this imported, ie. `use crate::vcs::git::Stats`

radicle-surf/src/error.rs
```diff
@@ -15,31 +15,21 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
-use radicle_surf::{file_system, git};
+//! error definitions for sources.
```
FintanH, line 18: Having this top-level seems weird to me because that would imply that it's a catch-all error type for all radicle-surf. Could you find where these error variants are used and localise them?

For example, say one of the errors is for constructing a `Commit`, then we would have another module `commit::error` that contains only those errors (not that this is the case but wanted to give an example to make it clearer :))

radicle-surf/src/object/blob.rs
```diff
@@ -15,26 +15,26 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
+//! Represents git object type 'blob'.
```
FintanH, line 18: I appreciate you adding the documentation :) It might be useful to add more text explaining what a blob is and perhaps linking to git documentation. If we're going to document we might as well be thorough :grin:

radicle-surf/src/oid.rs
```diff
@@ -15,12 +15,20 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
+//! Represents Oid in git
```
FintanH, line 18: I'd actually be more in favour of removing this and using `radicle_git_ext::Oid` instead, wdyt?

radicle-surf/src/revision.rs
```diff
@@ -83,18 +102,18 @@ where
...
     /// The peer peer_id for the user.
     pub peer_id: P,
     /// The user who owns these revisions.
     pub user: U,
-    /// List of [`git::Branch`].
-    pub branches: NonEmpty<Branch>,
-    /// List of [`git::Tag`].
-    pub tags: Vec<Tag>,
+    /// List of [`git::BranchName`].
+    pub branches: NonEmpty<BranchName>,
```
FintanH, line 114: What's the reason for the `BranchName` changes? Not that it's wrong, just curious as to why?

This might solve your issue about locality /cc @sebastinez

radicle-surf/src/person.rs
```diff
@@ -15,11 +15,15 @@
...
+
+#[cfg(feature = "serialize")]
 use serde::Serialize;
 
 /// Representation of a person (e.g. committer, author, signer) from a
 /// repository. Usually extracted from a signature.
-#[derive(Clone, Debug, Serialize)]
+#[cfg_attr(feature = "serialize", derive(Serialize))]
+#[derive(Clone, Debug)]
 pub struct Person {
```
FintanH, line 27: Can we create an issue for the person fields being `Option`. I don't think one _needs_ to set a name and email at the same time.

FintanH (commented):
Great start! :) I have some suggestions for improvements, let me know what you think!

radicle-surf/src/revision.rs
keepsimple1: The confusion is probably because the old comment was not correct, i.e. it was not `git::Branch`, it was `radicle_source::branch::Branch`, which is a branch name. 

Now we use `git::BranchName` instead.

radicle-surf/src/person.rs
keepsimple1: here you go: https://github.com/radicle-dev/radicle-git/issues/12

radicle-surf/src/error.rs
keepsimple1: Let me see if I can localize them. Eventually we might still need a crate level `Error` type for `radicle-surf` public API.

radicle-surf/src/oid.rs
keepsimple1: agreed to remove this. There is only one place in `revision.rs` using it, so I replaced it a `git2::Oid` for now. I will change it later to use `radicle_git_ext::Oid` when working on issue #5 & when I understand `radicle_git_ext` a bit better .

keepsimple1: I found that we cannot use plain `git2::Oid` as it does not support serialization. I went ahead to use `radicle_git_ext::Oid` ;-).

radicle-surf/src/revision.rs
```diff
@@ -15,45 +15,64 @@
...
+
 use std::convert::TryFrom;
 
 use nonempty::NonEmpty;
+
+#[cfg(feature = "serialize")]
 use serde::{Deserialize, Serialize};
 
-use radicle_surf::vcs::git::{self, Browser, RefScope, Rev};
+use radicle_git_ext::Oid;
```
keepsimple1, line 27: start to use `radicle_git_ext::Oid`.

radicle-surf/src/error.rs
FintanH: Possibly, but I'm more in favour of this idea of localising them and then building up to a smaller `Error` enum for the top-level concept. Similar to the idea of capabilities we chatted about on the call yesterday :)

Having a big `Error` enum with _all_ the variants becomes confusing when trying to reason about how your function fails. You can't tell in which ways a function can fail.

radicle-surf/src/object/blob.rs
FintanH: Ah yes, I'd actually be happy in turning that check off because it's very aggressive to have to document every single thing.

FintanH:
The code looks great! I have a nit about the commit message though :sweat_smile: 

> address review comments.

That means nothing when going back through a commit history. It's better to describe what the issue was, ie. a lot of duplication in source, and how it was fixed, ie. de-duplicated by merging into surf. For example, I would write something like:
```
radicle-surf: merge radicle-source into radicle-surf

Initially, radicle-source was envisioned to be another layer on top
of radicle-surf, however, it ended up duplicating a lot of concepts.

We consolidate them by merging radicle-source into radicle-source.
This meant removing modules <x>, <y>, and <z>.
```

I'd also prefer that the CI change was made into a separate commit if possible.
```
ci: fix issue with main branch

<again, the problem>

<again, the solution>
```

Sorry to be a stickler, but it will improve our lives when we need to go through git histories :) And definitely call me out on this when I'm not descriptive enough :wink:

radicle-surf/src/error.rs
keepsimple1: I'm on the same page. I didn't want a big `Error` enum especially at the beginning. Again in long term, we would need one at the top of the crate and should be kept small and defined carefully.

PR #11, opened 2022-09-20
Add `stats` method to `diff::Diff`

sebastinez:
Adds the `stats` method to `diff::Diff`, which calculates the amount of additions and deletions that happened in a Diff which is used in a few places in the `radicle-client-services/http-api`

### DIFF
--- a/radicle-surf/src/diff.rs
+++ b/radicle-surf/src/diff.rs
@@ -121,6 +121,22 @@ pub enum FileDiff {
     },
 }
 
+/// Statistics describing a particular [`Diff`].
+#[cfg_attr(
+    feature = "serialize",
+    derive(Serialize),
+    serde(rename_all = "camelCase")
+)]
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Stats {
+    /// Get the total number of files changed in a diff.
+    pub files_changed: usize,
+    /// Get the total number of insertions in a diff.
+    pub insertions: usize,
+    /// Get the total number of deletions in a diff.
+    pub deletions: usize,
+}
+
 /// A set of line changes.
 #[cfg_attr(
     feature = "serialize",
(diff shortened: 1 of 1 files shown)


radicle-surf/src/diff.rs
```diff
@@ -121,6 +121,20 @@ pub enum FileDiff {
...
 }
 
+/// Commit statistics.
+#[cfg_attr(
+    feature = "serialize",
+    derive(Serialize),
+    serde(rename_all = "camelCase")
+)]
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct CommitStats {
```
FintanH, line 131: I think this should just be called `Stats` it's not actually for a particular commit. Also, it'd be good to have the same fields as [`git2::DiffStats`](https://docs.rs/git2/0.15.0/git2/struct.DiffStats.html) -- so `files_changed` seems useful. Maybe we should reconsider keeping with the naming of `insertions` and `deletions`.

radicle-surf/src/diff.rs
```diff
@@ -121,6 +121,20 @@ pub enum FileDiff {
...
+    /// Additions.
```
FintanH, line 132: These docs are pretty useless :upside_down_face: It would be better to be more descriptive, e.g. "the total number of lines added in the [`Diff`]"

radicle-surf/src/diff.rs
```diff
@@ -121,6 +121,20 @@ pub enum FileDiff {
     },
 }
 
+/// Commit statistics.
```
FintanH, line 124: Also, this documentation should be changed:
```suggestion
/// Statistics describing a particular [`Diff`].
```

FintanH (commented):
Thanks for the change! Few suggestions :)

radicle-surf/src/diff.rs
sebastinez: Maybe we can call it even `DiffStats` instead of only `Stats`, else it doesn't really define what type of stats

radicle-surf/src/diff.rs
sebastinez: Done in [3c3527c](https://github.com/radicle-dev/radicle-git/pull/11/commits/3c3527c970b08e8c10f3b5944f7c6718f846d17b)

radicle-surf/src/diff.rs
sebastinez: Done in [3c3527c](https://github.com/radicle-dev/radicle-git/pull/11/commits/3c3527c970b08e8c10f3b5944f7c6718f846d17b)

FintanH:
You have some clippy errors :)

Could you also squash the commits, please? You could also provide more prose to the commit message describing the change :ok_hand:

sebastinez:
> You have some clippy errors :)
> 
> Could you also squash the commits, please? You could also provide more prose to the commit message describing the change 👌

Yeah no problem, had to go away quick 😅

sebastinez:
@FintanH Okay squashed and rewritten, ready for merge

FintanH:
Thanks!

PR #17, opened 2022-09-22
git-storage: port of librad storage

FintanH:
This provides a port of the [`librad::git::storage`](https://github.com/radicle-dev/radicle-link/blob/master/librad/src/git/storage.rs) API, with a focus on being more generic and decomposing writing and reading operations. It also takes heavy inspiration from [`link_replication::io::refdb`](https://github.com/radicle-dev/radicle-link/blob/master/link-replication/src/io/refdb.rs) for its implementation of the transactional writes.

It's currently a work in progress and hasn't been finalised but the essence is there. Some things left to do:
- [x] `Write` for `odb`
- [x] ~I'm tempted to split the `odb` capabilities for each object and then have a combination trait, ie. `trait Blob`, `trait Tree`, etc. `trait Odb: Blob + Tree + ...`. This would allow functions to say _which_ object they're accessing if they want to.~ This was too incoherent with naming and associated types.
- [x] ~Support optional namespacing for `backend::Write`~ Going to leave this for a follow-up
- [x] I'd like to go through the APIs and module/type naming and polish them up
- [x] Document everything properly


FintanH:
@cloudhead could you check this out it should be a suitable replacement for `librad::git::storage` and should have similar usage.

git-storage/src/config.rs
```diff
@@ -0,0 +1,225 @@
...
+            owner,
+            info,
+        };
+        this.guard_key_change()?;
+        this.ensure_reflog()?;
+        this.set_identifier(identifier)?;
+        this.set_user_info()?;
+
+        Ok(this)
+    }
```
cloudhead, line 174: Wondering if this still makes sense without a monorepo, ie. if we'll want to create this config in every project repo :thinking:, or have a global non-git config..

git-storage/src/backend/read.rs
```diff
@@ -0,0 +1,176 @@
...
+    type FindTree = git2::Error;
+
+    fn find_object(&self, oid: Oid) -> Result<Option<Object>, Self::FindObj> {
+        self.raw
+            .find_object(oid.into(), None)
+            .map(Some)
+            .or_matches::<git2::Error, _, _>(is_not_found_err, || Ok(None))
+    }
+
+    fn find_blob(&self, oid: Oid) -> Result<Option<Blob>, Self::FindBlob> {
```
cloudhead, line 149: I have mixed feelings on the `Result<Option<_>, _>` pattern, I guess because `None` can't be easily short-circuited, meaning I end up creating a `NotFound` error, which may as well be part of the `Result`. Perhaps the soon to come `let else` syntax will help with this?

cloudhead:
Looks nice and clean :+1:

keepsimple1 (dismissed):
Code looks good to me. Will there be a follow up PR for tests?

git-storage/src/config.rs
FintanH: Ya, I was thinking similar thoughts. On the same note, I was thinking that the signer/owner might not even make sense at this level. It might make more sense to push this a level up in radicle-node as we decide how things are organised, wdyt?

git-storage/src/backend/read.rs
FintanH: I equally have mixed feelings the other way, because then one may end up matching the `NotFound` error. I could add more fallible functions, eg.
```rust
pub fn blob<S: Read>(storage: S, oid: Oid) -> Result<Blob, S::FindBlob> {
    storage.find_blob(storage, oid).ok_or(git2::Error::new(
        git2::ErrorCode::NotFound,
        git2::ErrorClass:Object,
        format!("could not find blob {oid}")
    )
}
```

FintanH:
> Code looks good to me. Will there be a follow up PR for tests?

Ya, I can follow up with tests. I was getting fatigued working on it for a week :sweat_smile:

PR #18, opened 2022-09-22
radicle-surf: convert git2::Oid to radicle_git_ext::Oid

keepsimple1:
This is to fix issue #5 and also:

- fix CI errors on main branch.
~~- fix Rust 1.64.0 clippy warnings.~~ It was fixed by https://github.com/radicle-dev/radicle-git/pull/19

PR #19, opened 2022-09-23
nix: rust-1.64

FintanH:
rust-1.64 was announced[0].

Update the nix sources to pick up the latest stable. This update requires some fixes suggested by clippy.

[0]: https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html


FintanH:
@keepsimple1 I'd like to favour this PR instead of your commit https://github.com/radicle-dev/radicle-git/pull/18/commits/5ed9e0476b5192349a516a020237ca5acffc705f, just so that the nix update and the fixes are all in one :) Should be easy enough for you to drop the commit when you rebase :grin:

Issue #8, opened 2022-09-19
git-ext: add gen for Oid


FintanH:
Port over the following code to `radicle-git-ext/t`

```rust
use radicle_git_ext::{self as ext, Oid};

pub fn gen_oid(kind: git2::ObjectType) -> impl Strategy<Value = Oid> {
    any::<Vec<u8>>()
        .prop_map(move |bytes| git2::Oid::hash_object(kind, &bytes).map(Oid::from).unwrap())
}

pub fn gen_oid_with_zero(kind: git2::ObjectType) -> impl Strategy<Value = Oid> {
    prop_oneof![gen_oid(kind), Just(git2::Oid::zero().into()),]
}
```

FintanH:
Closed via #21

PR #21, opened 2022-09-27
radicle-git-ext/t: Port `gen_oid*` generator methods

jorgeig-space:
Addresses #8.

Ported from https://github.com/radicle-dev/radicle-link.

Note that the `Strategy` returns the radicle-git `Oid`type but it uses `git2::Oid` internally to generate it
since our `Oid` does not expose the `hash_object` and `zero` functions.

### DIFF
--- a/radicle-git-ext/t/Cargo.toml
+++ b/radicle-git-ext/t/Cargo.toml
@@ -14,13 +14,21 @@ doc = false
 [features]
 test = []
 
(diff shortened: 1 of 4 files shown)


FintanH (approved):
Nice! Thank you :)

Can you add new lines to the `gen.rs` and `urn`.rs` files? You can see GitHub calling it out with the red circle.

PR #23, opened 2022-09-27
radicle-surf: test structure compliance.

keepsimple1:
Fix issue #6 . 

The main goal is to setup the test structure for `radicle-surf` and collect existing tests. To allow all existing tests to run, we modify a couple of modules (`tree.rs`, `vcs::git::ext`) to be public, and a couple of `pub(crate)` functions to be `pub`. We can and will revisit them when refactoring `radicle-surf` crate further.

For `doctest`, I disabled it for `radicle-surf` crate, but leaving the existing doctest code as references.


deny.toml
```diff
@@ -180,8 +180,8 @@ deny = [
     # Each entry the name of a crate and a version range. If version is
     # not specified, all versions will be matched.
     #{ name = "ansi_term", version = "=0.11.0" },
-    { name = "openssl-probe" },
-    { name = "openssl-sys" },
+    { name = "openssl-probe", version = "<0.1.5" },
+    { name = "openssl-sys", version = "<0.9.76" },
```
keepsimple1, line 184: `cargo deny` reports errors for these two crates used by `git2 v0.15.0`.  This is a relative reasonable way I found to work around the problem.

deny.toml
```diff
@@ -180,8 +180,8 @@ deny = [
     # Each entry the name of a crate and a version range. If version is
     # not specified, all versions will be matched.
     #{ name = "ansi_term", version = "=0.11.0" },
-    { name = "openssl-probe" },
-    { name = "openssl-sys" },
+    { name = "openssl-probe", version = "<0.1.5" },
```
FintanH, line 183: You can revert this once you've done https://github.com/radicle-dev/radicle-git/pull/23/files#r982163684 :)

radicle-surf/src/vcs.rs
```diff
@@ -157,6 +157,11 @@ impl<Repo, A, Error> Browser<Repo, A, Error> {
         self.history.clone()
     }
 
+    /// Get the current `History` the `Browser` is viewing as a ref.
+    pub fn get_ref(&self) -> &History<A> {
```
FintanH, line 161: I think an `AsRef` impl might make more sense, or rename this `as_history`

radicle-surf/t/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+test = []
+
+[dev-dependencies]
+git2 = "0.15.0"
+nonempty = "0.5"
+pretty_assertions = "1.3.0"
+proptest = "0.9"
+serde_json = "1"
+
+[dev-dependencies.radicle-surf]
```
FintanH, line 23: nit: swap this with `radicle-git-ext` for alphabetical order :)

radicle-surf/t/src/lib.rs
```diff
@@ -0,0 +1,11 @@
+#[cfg(test)]
+mod test_git;
```
FintanH, line 2: I'd prefer without the `test_` prefix here. We know they're tests ;)

FintanH (commented):
Looking good! Few points of feedback :)

Could you also add the license header to new files:
```
// Copyright © 2022 The Radicle Git Contributors
// SPDX-License-Identifier: GPL-3.0-or-later
```

radicle-surf/src/vcs.rs
keepsimple1: Will opt for `as_history`.

radicle-surf/t/src/lib.rs
keepsimple1: For untrained eyes, files under `t/src/` are not obviously test files. Having `test_` prefix also helps in searching based on file names.

This is similar to have `test_` prefix in [test case names](https://github.com/radicle-dev/radicle-git/blob/d7b3da88a2ed2d59c170f02453bfd802f378d363/radicle-surf/src/diff.rs#L607).  There are pros and cons, but I think the pros overweight the cons.

FintanH: Just to give my further 2c on it, but not too pushed to have it changed either way.

> For untrained eyes, files under `t/src/` are not obviously test files. Having `test_` prefix also helps in searching based on file names.

I think the fact that we have documented that `t/src` are tests should be enough, also the crates are called `*-test` so adding to the evidence that they're test crates.

I would also argue that the `t` prefix is already enough to search based on file names :) But that all depends on the searching strategy right? :smile: My general philosophy is that the hierarchy gives way to denoting what components are and that prefixing is redundant :)

> This is similar to have `test_` prefix in [test case names](https://github.com/radicle-dev/radicle-git/blob/d7b3da88a2ed2d59c170f02453bfd802f378d363/radicle-surf/src/diff.rs#L607). There are pros and cons, but I think the pros outweigh the cons.

I think that was helpful in the case of the tests living inside the same module as functions, so no overlap when searching within a file, but now they live somewhere else. I wouldn't ask you to remove all the prefixes though because that would be cruel :joy:

keepsimple1: To move forward, I will remove `test_` prefix in this diff. Meantime, I will open an issue to track this discussion and we can reach a decision for the long term when solving that issue.

PR #24, opened 2022-09-29
[WIP] git-storage/t: Storage tests

jorgeig-space:
DRAFT

This PR adds tests for the `git-storage` functions in the `Read` and `Write` traits, implemented in the `Backend` struct.

So far, only one test to check whether the structure of the tests is as expected.


git-storage/t/src/properties/odb.rs
```diff
@@ -0,0 +1,31 @@
+// Copyright © 2022 The Radicle Link Contributors
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+use git_storage::{
+    odb::{Read, Write},
```
FintanH, line 5: You can import like this to avoid the name clashes :)
```suggestion
    odb::{Read as _, Write as _},
```

git-storage/t/src/properties/odb.rs
```diff
@@ -0,0 +1,31 @@
...
+use std::env::temp_dir;
+use uuid::Uuid;
+
+use crate::gen;
+
+proptest! {
+    #[test]
+    fn roundtrip_blob(user in gen::gen_signature(), data in gen::gen_bytes()) {
+        // Open the system's temp dir and use a distinct tmp directory per test
+        let mut dir = temp_dir();
```
FintanH, line 19: There's a [`tempfile`](https://docs.rs/tempfile/3.3.0/tempfile/) crate for this sort of thing.

In fact, we have helpers for creating temporary structures https://github.com/radicle-dev/radicle-git/blob/main/test/test-helpers/src/tempdir.rs

So you could define a `pub type TmpWrite = TmpDir<Write>` to create temporary backends.

FintanH (commented):
First test is looking good :grin:

git-storage/t/src/properties/odb.rs
jorgeig-space: Oh cool! Ok, I'll use that and create a function in helpers.rs to open the repo.

jorgeig-space:
I'm thinking about removing the `*_object` tests and just incorporating them in the normal roundtrip tests, like in the `roundtrip_tag` function. That means each roundtrip test would be testing two things, but it also makes the test code much shorter since there's no need to duplicate creating trees/commits. What do you think @FintanH?

FintanH (commented):
I think this is a good start! I think that the next phase could try writing generators for the objects themselves, which will be a bit trickier.

I think we should start with `gen_blob` and work from there. So there's nothing stopping you from writing something like:
```rust
pub fn gen_blob(storage: &Write, bytes: &[u8]) -> impl Strategy<Value = Blob> {
  todo!()
}
```

From there we can make files and I'll change what I said in another comment:
```rust
pub struct File {
  path: PathBuf,
  inner: Blob,
  mode: git2::FileMode
}

pub fn gen_file(storage: Write) -> impl Strategy<Value = File> {
  gen_path().prop_map(|path| 
    gen_bytes().prop_map(|bytes|
      gen_blob(storage, &bytes).prop_map(|blob|
        gen_mode().map(|mode| File { path, inner: blob, mode }))))
}
```

We can then do  `Tree`:
```rust
pub fn gen_tree(storage: &Write) -> impl Strategy<Value = Tree> {
  /* a little bit more complex but we can generate a vector of blobs and insert them into a TreeBuilder */
}
```

Then we do a `Commit` and then we do `Tag`.

Does that make sense?

git-storage/t/src/properties/odb.rs
```diff
@@ -0,0 +1,162 @@
...
+        let path = repo.workdir().unwrap();
+
+        build_tree_from_files(files, path, &mut [&mut tree_builder], oid, mode);
+
+        let writer = Write::open(path, user).unwrap();
+        let tree_oid = writer.write_tree(tree_builder).unwrap();
+        let tree = writer.read_only().find_tree(tree_oid).unwrap().unwrap();
+        let commit_oid = writer.write_commit(&tree, &[], &message).unwrap();
+        let readback = writer.read_only().find_commit(commit_oid).unwrap().unwrap();
+        assert_eq!(readback.tree_id(), tree_oid.into());
```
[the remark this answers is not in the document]
jorgeig-space: I'll put a comment there. I actually was checking that the associated tree had the ID of the tree I specified in the write. But it's also true that we can add a check for the actual commit ID (and the same is true for other tests).

git-storage/t/src/properties/odb.rs
```diff
@@ -0,0 +1,162 @@
...
+        // a `Tree`. Writing again the `TreeBuilder` and comparing that three is meaningless,
+        // so here we walk the read Tree and try to find each node in the TreeBuilder
+        readback_tree.walk(TreeWalkMode::PreOrder, |_root, entry| {
+            let _expected_entry = tree_expected.get(entry.name().unwrap()).unwrap().unwrap();
+            TreeWalkResult::Ok
+        }).unwrap();
+    }
+
+    #[test]
+    fn roundtrip_commit(user in gen::gen_signature(), files in gen::gen_files(), mode in gen::gen_valid_mode(), oid in gen_oid(git2::ObjectType::Commit), message in gen::trivial()) {
```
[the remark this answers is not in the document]
jorgeig-space: Ah I was wondering why rustfmt didn't do that, and it's because of the macro. Duh! Yeah no worries I'll do that manually.

git-storage/t/src/gen.rs
```diff
@@ -0,0 +1,42 @@
...
+            email: format!("{}@{}.com", &name, &name),
+        }
+    })
+}
+
+pub fn gen_bytes() -> impl Strategy<Value = Vec<u8>> {
+    any::<Vec<u8>>()
+}
+
+pub fn gen_files() -> impl Strategy<Value = Vec<(String, Vec<u8>)>> {
```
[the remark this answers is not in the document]
jorgeig-space: Agreed.

git-storage/t/src/properties/odb.rs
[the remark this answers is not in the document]
jorgeig-space: Actually, checking the Oids is enough, so no need to do this. I've removed those additional checks.

jorgeig-space:
Thanks for your help @FintanH. I think after some cleanup this is ready to merge. Will open another PR for the `refdb` side.

git-storage/t/src/helpers.rs
```diff
@@ -0,0 +1,22 @@
+use git2::Repository;
+use git_storage::{signature::UserInfo, Write};
+use test_helpers::tempdir::WithTmpDir;
+
+pub type TmpWriter = WithTmpDir<Write>;
+pub type TmpRepo = WithTmpDir<Repository>;
+
+pub fn new_tmp_writer(user: UserInfo) -> TmpWriter {
```
FintanH, line 8: nit: can we call this module `tmp`, call this function `writer`, and use as `tmp::writer`?

git-storage/t/src/helpers.rs
```diff
@@ -0,0 +1,22 @@
...
+    WithTmpDir::new(|path| {
+        Write::open(path, user)
+            .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, format!("{}", e)))
+    })
+    .unwrap()
+}
+
+pub fn new_tmp_repo() -> TmpRepo {
```
FintanH, line 16: Is this still used or can we remove it?

test/test-helpers/src/tempdir.rs
```diff
@@ -11,6 +11,7 @@ use std::{
 
 use tempfile::{tempdir, TempDir};
 
+#[derive(Debug)]
```
FintanH, line 14: Is this needed?

git-storage/src/backend/read.rs
```diff
@@ -46,6 +46,14 @@ impl Read {
     }
 }
 
+impl core::fmt::Debug for Read {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Read")
+            .field("raw", &self.raw.path())
+            .finish()
+    }
+}
```
FintanH, line 55: nit: I prefer importing `std::fmt`
```suggestion
impl fmt::Debug for Read {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        f.debug_struct("Read")
            .field("raw", &self.raw.path())
            .finish()
    }
}
```

FintanH (commented):
Looks great! Few suggestions :)

test/test-helpers/src/tempdir.rs
jorgeig-space: Only when the repo generator was used, not it's not needed.

git-storage/t/src/helpers.rs
jorgeig-space: It can be removed.

git-storage/src/backend/read.rs
jorgeig-space: Ah sure, force of habit for no-std!

FintanH (approved):
Beautiful!

FintanH (approved):

PR #26, opened 2022-09-29
radicle-surf: refactoring the design - part 1

keepsimple1:
This is for issue #27 .

This is the first patch to refactor the design of `radicle-surf`. This patch includes:

- A new design document with the first set of changes proposed. ([rendered](https://github.com/keepsimple1/radicle-git/blob/new-design/radicle-surf/docs/refactor-design.md))
- Deleted `Browser` and `Snapshot`, as described in the design doc, and implemented in code changes. The main functions are implemented using `RepositoryRef`.


radicle-surf/docs/refactor-design.md
```diff
@@ -0,0 +1,46 @@
+# An updated design for radicle-surf
```
FintanH, line 1: I think I agree with the essence of the document :)

I reckon we can structure it something like:
```
## Introduction

<explain what surf is meant to achieve>

## Motivation

<explain why we're refactoring>

## Review

<review the existing types and say why or why not we should keep them, or something like them>

## High-Level API

<this is where the document gets interesting, what kind of functionality do we want from our API>
<use GitHub as a source of inspiration>
<talk to UI clients people for understanding what they need>
<anticipate stuff coming down the line, for example, should a user of this library be able to extend it to browse issues via cob refs>

### Types (or Components)

<this could be optional, maybe we want to do this in another spec>
<describe the high-level types and their semantics>
```

cloudhead:
When considering a new design, I have just two additional notes:
* Browser is currently inefficient for retrieving HEAD, or commit ranges, as it always materializes the full commit history when trying to fetch a commit.
* Browser is currently "stateful", ie. you set it to a certain state, then query it. I'm not sure that design buys us anything, and I think it would be nicer to have a more stateless API at the top level, with perhaps stateful sub-components (iterators, remotes etc.).

radicle-surf/docs/refactor-design.md
keepsimple1: Thanks, I will incorporate this in the next iteration of the doc.

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -171,14 +179,90 @@ impl<'a> RepositoryRef<'a> {
         Ok(self.repo_ref.revparse_single(oid)?.id().into())
     }
 
+    /// Gets a snapshot of the repo as a Directory.
+    pub fn snapshot(&self, rev: &Rev) -> Result<directory::Directory, Error> {
```
keepsimple1, line 183: This is one of main methods in the API.

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -171,14 +179,90 @@ impl<'a> RepositoryRef<'a> {
...
+        let commit = self.rev_to_commit(rev)?;
+        let oid: Oid = commit.id().into();
+        let tree = self.get_tree(&oid)?;
+        Ok(directory::Directory::from_hash_map(tree))
+    }
+
+    /// Returns the last commit, if exists, for a `path` in the history of
+    /// `rev`.
+    pub fn last_commit(&self, path: file_system::Path, rev: &Rev) -> Result<Option<Commit>, Error> {
```
keepsimple1, line 192: A different design is to compose `RepositoryRef` with `History`, e.g. the user can call:
```
repo.get_history(rev).last_commit(path)
```

For the 1st iteration, I opted to use the current one to take advantage of existing supporting functions.

FintanH (approved):
Happy to merge this as a first iteration :)

Issue #27, opened 2022-09-29
radicle-surf: refactoring the design


keepsimple1:
There been a desire to come up with a new / refactored design for `radicle-surf`. This issue is to track that effort, including:

- A design document.
- An implementation of the design.

FintanH:
I've been trying to think about this redesign a bit. I believe there are essentially 4 parts to it:
1. refdb (reference database) listing
2. odb (object database) querying
3. diff'ing
4. code browsing
5. other metadata

## 1. Refdb

This component is all about listing/counting & querying for different types of references. Since we're in the domain of git code browsing we can also specialise these to the known categories:
* `heads`
* `tags`
* `remotes`
* `notes`

But being good citizens, we can leave the door open for a supplied `refs/<category>`.

Being able to query a reference for its `Oid` will be necessary for then plugging into the `Odb` component since that's how objects are found.

## 2. Odb

We will want to be able to query for objects themselves:
* `Commit`
* `Tag`
* `Tree`
* `Blob`

On top of this, we will want to be able to wrap around a `revwalk`, which would help with this issue of loading all of the commit history into memory. It will only need a single `Oid` and the interface is an iterator.

Instead of our own non-empty `Tree` implementation, we can probably just adapt around the `git2::Tree` interface and make it nicer to use.

## 3. Diff'ing

Similar to `git2::Tree`, I think we could wrap around `git2::Diff` to have a nicer interface as opposed to having our own diff'ing implementation.

Another alternative would be to use something like `difftastic`.

## 4. Code browsing

This is essentially a combination of Refdb + Odb. The input for its actions will be something that resolves to a `Commit`'s `Oid`. Some actions could:
* Get the commit history - takes the `Oid` and gives you back the revwalk
* Get the directory view - takes the `Oid` and materialises the `Tree`

There's probably more, but we can add those as we go :)

## 5. Metadata

This is more of a miscellaneous category such as contributor statistics, for example. I don't have an exhaustive list for these either.

Wdyt? Would be interested to hear if you wanted to design anything differently or if I'm missing anything :)

FintanH:
Oh, and something I forgot to mention is that the refdb + odb work is pretty much done in `git-storage` and we could build specialised functions on top of those lower-level abilities.

cloudhead:
Makes a lot of sense. One thing I'd like to understand better though is: which of the following are we doing, and to what extent:
1) Adding general git functionality not available in git2
[...]

PR #28, opened 2022-10-07
refactor design: part 2

keepsimple1:
Continue to work on issue #27. Part 2:

- Update design doc, covering "code browsing" and "diff".
- Add a code browsing test for new API.
- Simplify `Directory`.
- Remove `Tree` and `Forest`.
- Update Diff API.


radicle-surf/docs/refactor-design.md
```diff
@@ -1,19 +1,26 @@
...
+## Introduction
 
-- Given a commit (or other types of ref), list the content: i.e. files and directories.
-- Generate a diff between two commits.
-- List the history of the commits.
-- List refs: Branches, Tags.
+Now we have ported the `radicle-surf` crate from its own github repo to be part of the `radicle-git` repo. We are taking this opportunity to refactor its design as well. Intuitively, `radicle-surf` provides an API so that one can use it to create a GitHub-like UI for a git repo:
 
-The main goals of the changes are:
+1. Code browsing: given a specific commit/ref, browse files and directories.
```
FintanH, line 7: I would say we would want to consider `commit/ref` as `revision`[[0]]. Admittedly, revisions are complex and I'm not even sure we'd wanna support things like commit ranges (that would be more for commit histories).

Another way I'd think about this is the input should be something that "peels"/"resolves" to a commit's `Oid`. So one potential route could be:

```rust
pub trait Peeled<Storage> {
    type Error;
    
    fn into_oid(&self, storage: &Storage) -> Result<Oid, Self::Error>;
}

impl<S> Peeled<S> for Oid {
    type Error = Infallible;
    
    fn into_oid(&self, _: &Storage) -> Result<Oid, Self::Error> {
        Ok(*self)
    }
}

impl<S: refdb::Read> Peeled<S> for RefStr {
    type Error = refdb::FindError;
    
    fn into_oid(&self, storage: &Storage) -> Result<Oid, Self::Error> {
        Ok(storage.find_reference(self)?.target)
    }
}

pub fn snapshot<S, P>(storage: &S, peeled: P) -> Result<Directory, Error>
where
    S: refdb::Read + odb::Read,
    P: Peeled<S>,
{
    let oid = peeled.into_oid(storage)?
    Ok(directory(storage, oid)?)
}
```

Sorry, that was a bit of a tangent based on "commit/ref" :joy: Wdyt? :)

[0]: https://git-scm.com/docs/gitrevisions

radicle-surf/docs/refactor-design.md
```diff
@@ -1,19 +1,26 @@
 # An updated design for radicle-surf
 
-Now we have ported the `radicle-surf` crate from its own github repo to be part of the `radicle-git` repo. We are taking this opportunity to refactor its design as well. Intuitively, `radicle-surf` provides an API so that one can use it to create a github-like UI for a git repo:
+## Introduction
 
-- Given a commit (or other types of ref), list the content: i.e. files and directories.
-- Generate a diff between two commits.
-- List the history of the commits.
-- List refs: Branches, Tags.
+Now we have ported the `radicle-surf` crate from its own github repo to be part of the `radicle-git` repo. We are taking this opportunity to refactor its design as well. Intuitively, `radicle-surf` provides an API so that one can use it to create a GitHub-like UI for a git repo:
```
FintanH, line 5: nit: can we do 80 wide column formatting for docs, plz? :pray:

radicle-surf/docs/refactor-design.md
```diff
@@ -1,19 +1,26 @@
...
 
-The main goals of the changes are:
+1. Code browsing: given a specific commit/ref, browse files and directories.
+2. Diff between two commits.
```
FintanH, line 8: Again, the high-level operation could be two things that resolve into two commits. Low-level would be the diff between two commits.

radicle-surf/docs/refactor-design.md
```diff
@@ -1,19 +1,26 @@
...
+3. Retrieve the history of the commits.
+4. Retrieve a specific commit: all its metadata.
```
FintanH, line 10: I would extend this to all objects.

radicle-surf/docs/refactor-design.md
```diff
@@ -1,19 +1,26 @@
...
+5. Retrieve the refs: Branches, Tags, Remotes.
```
FintanH, line 11: I would extend this to include "Notes" and also open it for user-defined "categories", where a category is:
`refs/<category>/<rest>`

radicle-surf/docs/refactor-design.md
```diff
@@ -1,19 +1,26 @@
...
+## Motivation
 
-## Make API simpler
+The `radicle-surf` crate aims to provide a safe and easy-to-use API that supports the features listed in [Introduction]. Based on the existing API, the main goals of the refactoring are:
+
+- API review: make API simpler whenever possible.
+- Address open issues in the original `radicle-surf` repo as much as possible.
+- Not to shy away from being `git` specific. (i.e. not to consider supporting other VCS systems)
+- Hide away `git2` from the API. The use of `git2` should be an implementation detail.
+- Make sure we support `rad://` references.
```
FintanH, line 21: This doesn't make much sense. The `rad://` is the start of a URL. Supporting rad references would be `refs/rad` but there will also be `refs/cob`. I mentioned this above a user-defined "category".

radicle-surf/docs/refactor-design.md
keepsimple1: Will update.

radicle-surf/docs/refactor-design.md
keepsimple1: My bad. Will update.

radicle-surf/docs/refactor-design.md
keepsimple1: OK.

radicle-surf/docs/refactor-design.md
keepsimple1: Will do.

radicle-surf/docs/refactor-design.md
FintanH: > I'm not sure about the version using generics, and taking the extra parameter of storage

Can you explain why? I'm of the opinion that we shouldn't use `RepositoryRef` but would like to hear your issues with the above first :)

radicle-surf/t/src/git.rs
```diff
@@ -784,3 +784,34 @@ mod reference {
...
+        let root_dir = repo.snapshot(&Branch::local("master").into()).unwrap();
+        let count = println_dir(&root_dir, 0);
+        assert_eq!(count, 36); // Check total file count.
+
+        /// Prints items in `dir` with `indent_level`.
+        /// For sub-directories, will do Depth-First-Search and print
+        /// recursively.
+        fn println_dir(dir: &Directory, indent_level: usize) -> i32 {
+            let mut count = 0;
+            for item in dir.iter() {
```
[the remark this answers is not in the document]
keepsimple1: With simplified `Directory`, now `iter()` returns a reference.

radicle-surf/src/file_system/directory.rs
```diff
@@ -308,25 +292,8 @@ impl Directory {
...
-                        .map(|sub_tree| sub_tree.into()),
-                )
-            },
-        }
-
-        empty_iter
-            .into_iter()
-            .flatten()
-            .chain(trees_iter.into_iter().flatten())
+    pub fn iter(&self) -> impl Iterator<Item = &DirectoryContents> {
```
keepsimple1, line 295: The `Item` returned is now a reference. No more clone.

radicle-surf/t/src/file_system.rs
```diff
@@ -89,113 +89,6 @@ mod directory_size {
...
-
-    // TODO(fintan): This is a bit slow. Could be time to benchmark some functions.
-    proptest! {
-        #[test]
-        fn prop_test_all_directories_and_files(directory_map in directory_map_strategy(10, 10, 10)) {
-            prop_all_directories_and_files(directory_map);
-        }
-    }
-
-    fn prop_all_directories_and_files(
```
keepsimple1, line 157: This is removed because we no longer need `Directory::from_hash_map` method. If anyone wants to transform these tests to verify new methods / functions, please let me know and I can try.

radicle-surf/src/file_system/directory.rs
```diff
@@ -464,17 +448,50 @@ impl Directory {
...
-        self.sub_directories.insert(path.0, file)
+        let name = path.0.last().clone();
+        let label = name.clone();
+        let f = DirectoryContents::File { name, file };
+
+        let mut contents = &mut self.contents;
+        let path_depth = path.0.len() - 1; // exclude the last label: file name
+
+        if path_depth == 0 {
+            contents.insert(label, f);
```
FintanH, line 473: nit
```suggestion
        let f = DirectoryContents::File { name: name.clone(), file };

        let mut contents = &mut self.contents;
        let path_depth = path.0.len() - 1; // exclude the last label: file name

        if path_depth == 0 {
            contents.insert(name, f);
```

radicle-surf/src/file_system/directory.rs
```diff
@@ -464,17 +448,50 @@ impl Directory {
...
+        } else {
+            for (idx, label) in path.iter().enumerate() {
+                // if label does not exist, create a sub directory.
+                if contents.get(label).is_none() {
```
FintanH, line 477: I feel like it should be possible to combine this `if` with the `match` below. Something like:
```rust
match contents.get_mut(label) {
  None => { /* do the insert & update contents pointer */ },
  Some(DirectoryContents::Directory(d)) => {
    /* code */
  },
  Some(DirectoryContents::File { .. }) => return,
}
```

I could be wrong though, so perhaps you could try to confirm? It would look a lot cleaner :)

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -441,97 +441,99 @@ impl<'a> RepositoryRef<'a> {
...
-    ) {
-        files
-            .entry(path)
-            .and_modify(|entries| entries.push((name.clone(), file.clone())))
-            .or_insert_with(|| NonEmpty::new((name, file)));
-    }
+    /// Generates a Directory for the commit.
+    fn directory_of_commit(&self, commit: &git2::Commit) -> Result<Directory, Error> {
+        let mut parent_dirs = vec![Directory::root()];
+        let git2_tree = commit.as_object().peel_to_tree()?;
```
FintanH, line 447: nit: we don't need the extra `git2` naming :) Editors can show the type if someone needs to know where it comes from
```suggestion
        let tree = commit.as_object().peel_to_tree()?;
```

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -441,97 +441,99 @@ impl<'a> RepositoryRef<'a> {
...
-                    match file_paths_or_error.as_mut() {
-                        Ok(files) => Self::update_file_map(path, name, file, files),
-
-                        // We don't need to update, we want to keep the error.
-                        Err(_err) => {},
+            match entry.kind() {
+                Some(git2::ObjectType::Tree) => {
+                    // Add a new level of directory.
+                    if let Some(name) = entry.name() {
+                        match Label::from_str(name) {
```
FintanH, line 468: fyi: you can also use `name.parse()`, or `name.parse::<Label>()` if it can't figure out the type.

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -37,12 +37,14 @@ use crate::{
...
 };
+use directory::Directory;
 use nonempty::NonEmpty;
 use radicle_git_ext::Oid;
 use std::{
-    collections::{BTreeSet, HashMap, HashSet},
+    collections::{BTreeSet, HashSet},
     convert::TryFrom,
     str,
+    str::FromStr,
```
FintanH, line 47: I tend to try and import traits with an `_`, so `use str::FromStr as _` -- that is if they're only being used to bring the methods into scope. It's a nice indication that we're not defining a `FromStr` in this module.

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -441,97 +441,99 @@ impl<'a> RepositoryRef<'a> {
...
-                        // We don't need to update, we want to keep the error.
-                        Err(_err) => {},
+            match entry.kind() {
+                Some(git2::ObjectType::Tree) => {
+                    // Add a new level of directory.
+                    if let Some(name) = entry.name() {
+                        match Label::from_str(name) {
+                            Ok(label) => parent_dirs.push(Directory::new(label)),
+                            Err(e) => {
+                                println!("label from {}: {}", name, &e);
```
FintanH, line 471: Can we remove this `println`. If you'd like to introduce logging then we should use the `log` crate.

Also, I did prefer that we were returning an error from the original function.

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -441,97 +441,99 @@ impl<'a> RepositoryRef<'a> {
...
-                    TreeWalkError::Git(err) => {
-                        file_paths_or_error = Err(err);
-                        git2::TreeWalkResult::Abort
-                    },
+                Some(git2::ObjectType::Blob) => {
+                    // Construct a File to insert into its parent directory.
+                    let object = match entry.to_object(self.repo_ref) {
+                        Ok(obj) => obj,
+                        Err(e) => {
+                            println!("entry to_object: {}", &e);
```
FintanH, line 482: Same here about printing and returning an error

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -441,97 +441,99 @@ impl<'a> RepositoryRef<'a> {
...
+                            return git2::TreeWalkResult::Abort;
+                        },
+                    };
+                    let blob = match object.as_blob() {
+                        Some(b) => b,
+                        None => return git2::TreeWalkResult::Abort,
+                    };
```
FintanH, line 489:
```suggestion
                    let blob = match entry.to_object(self.repo_ref).and_then(|obj| obj.as_blob()) {
                        Ok(b) => b,
                        Err(e) => {
                            println!("entry to_object: {}", &e);
                            return git2::TreeWalkResult::Abort;
                        },
                    };
```

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -441,97 +441,99 @@ impl<'a> RepositoryRef<'a> {
...
+                if let Some(parent) = parent_dirs.last_mut() {
+                    let name = last_dir.name().clone();
+                    let content = DirectoryContents::Directory(last_dir);
+                    parent.insert(name, content);
+                } else {
+                    break last_dir;
+                }
+            } else {
+                // Fail-safe: create an empty directory.
+                break Directory::root();
```
FintanH, line 532: Mmmmm I'm not so sure about this, it smells to me. Can you explain what's going on in this loop? My feeling is that you might be able to better express the logic using a `NonEmpty` vector or something.

radicle-surf/src/file_system/directory.rs
```diff
@@ -258,18 +247,13 @@ impl Directory {
...
-            Some(trees) => trees
-                .0
-                .iter()
-                .map(|tree| match tree {
-                    SubTree::Node { key: name, .. } => SystemType::file(name.clone()),
-                    SubTree::Branch { key: name, .. } => SystemType::directory(name.clone()),
-                })
-                .collect(),
-        }
+        self.contents
```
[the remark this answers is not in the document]
keepsimple1: In fact, the current `iter()` (or `contents()` as the name you suggested) is doing that. I don't have problems removing this method as it's overlapping.

radicle-surf/src/file_system/directory.rs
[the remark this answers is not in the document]
keepsimple1: I am OK with either one, and slightly prefer `iter()` because of the return type `impl Iterator<>`.

radicle-surf/src/file_system/directory.rs
keepsimple1: It has pros and cons. The cons is that, `None` branch would also check for `if idx +1 == path_depth`, i.e. if we are in the last directory level, and if yes, would insert the actual file.   That will be duplicating the logic in `Some(DirectoryContents::Directory(d))` branch.   

The current approach does not duplicate the logic by doing lookup twice (only when the path label did not exist). I opt to keep this approach for now.

radicle-surf/src/vcs/git/repo.rs
keepsimple1: Sorry about the confusion. I updated the comments and the code. Hope it is clearer.

radicle-surf/src/vcs/git/repo.rs
keepsimple1: However,  `obj.as_blob()` returns `Option`, not `Result`.  If I understand correctly, [ `.and_then()`](https://doc.rust-lang.org/std/result/enum.Result.html#method.and_then) requires a function that returns `Result`.

radicle-surf/src/file_system/directory.rs
keepsimple1: done.

radicle-surf/src/file_system/directory.rs
```diff
@@ -434,10 +421,7 @@ impl Directory {
     /// assert_eq!(test.current(), unsound::label::new("test"));
     /// ```
     pub fn current(&self) -> Label {
-        match &self.current {
-            Location::Root => Label::root(),
-            Location::SubDirectory(label) => label.clone(),
-        }
+        self.name.clone()
```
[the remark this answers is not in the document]
keepsimple1: done.

radicle-surf/src/file_system/directory.rs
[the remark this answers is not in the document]
FintanH: Ya, let's de-duplicate where we can :)

radicle-surf/src/file_system/directory.rs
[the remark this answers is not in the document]
FintanH: I suggested `contents` because of the return type of the `Item` :) Similar to how `BTreeMap` has `keys` and `values`. I could foresee having another iterator:
```rust
pub fn files(&self) -> Files<'a>
```
which returns an `Iterator<Item = &File>` :)

radicle-surf/src/file_system/directory.rs
FintanH: > The current approach avoids duplicating the logic by doing lookup twice. I opt to keep this approach for now.

Out of curiosity, why do you favour doing a look-up twice over repeating an `if` statement? :thinking: I would think that a look-up would be more expensive in comparison to the addition and compare. Basically, I'm curious as to how you're evaluating one is better of the other :)

radicle-surf/src/vcs/git/repo.rs
FintanH: Whoops! Should have looked at the docs :) Same principle, but using `map`:
```rust
match entry.to_object(self.repo_ref).map(|obj| obj.as_blob()) {
  Ok(Some(b)) => /* code */
  Ok(None) => /* code */,
  Err(e) => /* code */,
}

File: radicle-surf/src/file_system/directory.rs [hunk 21]
```diff
@@ -165,50 +159,45 @@ pub enum DirectoryContents {
...
+    pub fn new(name: Label) -> Self {
         Directory {
-            current: Location::SubDirectory(label),
-            sub_directories: Forest::root(),
+            name,
+            contents: BTreeMap::new(),
         }
     }
 
+    /// get the name
```
FintanH, line 191: nit: let's follow English grammar rules and start with a capital :)
```suggestion
    /// Get the name of the current `Directory`.
```

radicle-surf/src/file_system/directory.rs
```diff
@@ -165,50 +159,45 @@ pub enum DirectoryContents {
...
+    pub fn name(&self) -> &Label {
+        &self.name
+    }
+
+    /// add an entry to the directory.
```
FintanH, line 196:
```suggestion
    /// Add the `content` under `name` to the current `Directory`.
    /// If `name` already exists in this directory, then the previous contents are replaced.
```

radicle-surf/t/src/git.rs
```diff
@@ -784,3 +784,50 @@ mod reference {
...
+                    count += println_dir(sub_dir, indent_level + 1);
+                }
+            }
+            count
+        }
+    }
+
+    // Show that list_directory() overlaps with iter().
+    #[test]
+    fn list_directory_nested() {
```
FintanH, line 819: I'd expect this to assert something but it's just printing?

radicle-surf/src/vcs/git/repo.rs
keepsimple1: It's also trickier than I thought: 

```
error[E0515]: cannot return reference to function parameter `obj`
   --> radicle-surf/src/vcs/git/repo.rs:477:79
    |
477 |                     let blob = match entry.to_object(self.repo_ref).map(|obj| obj.as_blob()) {
    |                                                                               ^^^^^^^^^^^^^ returns a reference to data owned by the current function

For more information about this error, try `rustc --explain E0515`.
```

There might be some clever tricks (I don't know) that can help chaining these two calls together, but for now I will focus on other changes and keep the current straightforward approach.

radicle-surf/t/src/git.rs
keepsimple1: I was only trying to test the output of `list_directory()` and compare it with `iter()`.  In the next version, I will de-dup them so `list_directory` will mostly be gone and this test case will no longer be needed.

radicle-surf/src/file_system/directory.rs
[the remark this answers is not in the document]
keepsimple1: I will change `iter()` to `contents()`.

radicle-surf/src/file_system/directory.rs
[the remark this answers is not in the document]
keepsimple1: will remove both `list_directory()` and `SystemType`.

radicle-surf/docs/refactor-design.md
keepsimple1: I'm not convinced yet that we should give up `RepositoryRef` (and `Repository` type). I like to postpone this possible change to the next (or future) diff.

radicle-surf/src/vcs/git/repo.rs
FintanH: Ya, fair enough. Rust's ownership is a PITA sometimes :upside_down_face:

radicle-surf/docs/refactor-design.md
```diff
@@ -1,47 +1,153 @@
...
+its design as well. Intuitively, `radicle-surf` provides an API so that one
+can use it to create a GitHub-like UI for a git repo:
 
-The main goals of the changes are:
+1. Code browsing: given a specific commit/ref, browse files and directories.
+2. Diff between two revisions that resolve into two commits.
+3. Retrieve the history of the commits.
+4. Retrieve a specific object: all its metadata.
+5. Retrieve the refs: Branches, Tags, Remotes, Notes and user-defined
+"categories", where a category is: refs/<category>/<...>.
```
FintanH, line 15: This is more of a note to all of us. Git tends to call these namespaces -- which _may_ be confusing compared to gitnamespaces. But wanted to make a note of it because I've read it in their docs a few times now and it might be good to be consistent with them :)

radicle-surf/docs/refactor-design.md
```diff
@@ -1,47 +1,153 @@
...
+In our API, it will be good to have a single way to identify all refs and
+objects that resolve into commits. In other words, we try to avoid using
+different ways at different places. Currently there are multiple types in the
+API for this purpose:
+
+- Commit
+- Oid
+- Rev
+
+Because `Rev` is the most high level among those and supports refs already,
```
FintanH, line 89: Feel free to remove/redefine types too :) I will not be precious about this code and mandate we need to keep any of it :smile:

FintanH (approved):
Sweet :ok_hand:  Nice work @keepsimple1! Let's merge and keep this train moving :)

PR #31, opened 2022-10-12
git-ref-format: qualified macro

FintanH:
Add a `qualified!` macro to the set of git-ref-format macros. This allows the construction of a `refs/<cat>/<name>` reference from a string literal.


cloudhead (approved):
Ah this is great!

Issue #15, opened 2022-09-21
surf: convert to git-ref-format


FintanH:
`radicle-surf`, historically, defined its own domain of reference types. This has, historically, caused some headaches for converting from surf types into `radicle-link` types. Now that `git-ref-format` is available to surf, we should plan out the conversion of the API to use those types instead.

- [ ] Look at the current API and where it uses reference types
- [ ] Plan out a migration path based on the above

Some things to consider:
* Whether the input should always be `Qualified`, ie. have a category like `refs/heads`, `refs/tags`, `refs/remotes`
* If we should do any newtype'ing to distinguish qualified refs, e.g. `pub struct Tag(Qualified)`, with smart constructors
* Are namespaces important anymore if we go down the protocol route of having a single repo per project? Thus, keep namespaces in the API but can de-prioritise the implementation of them

cloudhead:
Good point about namespaces; we aren't using them in the new protocol!

keepsimple1:
When I was working on the PR, I was wondering what are the best examples of `git-ref-format`'s benefits?  I understand the points in its doc page, but it is not clear to me how much benefits it brings to overweight the cost.

FintanH:
> When I was working on the PR, I was wondering what are the best examples of `git-ref-format`'s benefits? I understand the points in its doc page, but it is not clear to me how much benefits it brings to overweigh the cost.

What are the costs you refer to here?

The benefits are:
* It's a well-typed interface. It's useful to constrain your domain to well-known types, i.e. `RefStr`/`RefString` are valid references, `Qualified` is a valid reference that begins with `refs/<category>`, `Namespaced` is a valid reference that begins with `refs/namespaces`, and `PatternString`/`PatternStr` are valid refspecs. This is opposed to haphazardly passing around strings that we hope are correct. Another way of thinking about this is, ["Parse, Don't Validate"](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/)
* We use `git-ref-format` in important places such as [`git-storage`](https://github.com/radicle-dev/radicle-git/tree/main/git-storage), [`radicle-cob`](https://github.com/radicle-dev/heartwood/blob/master/radicle-cob/Cargo.toml#L17), and [`radicle`](https://github.com/radicle-dev/heartwood/blob/master/radicle/Cargo.toml#L19). That means that those libraries will already passing around and returning types from `git-ref-format`.

keepsimple1:
[...]

PR #37, opened 2022-10-21
radicle-surf: refactor design: part 4

keepsimple1:
Continued work for https://github.com/radicle-dev/radicle-git/issues/27, part 4:

- Update design doc.
- Refactor History API.
- Use `FromRevision` and `IntoCommit` following suggestions in https://github.com/radicle-dev/radicle-git/pull/38
- Add more test cases.


radicle-surf/docs/refactor-design.md
```diff
@@ -143,17 +145,37 @@ The core API is:
...
+`RepositoryRef`.
+
 ```Rust
 imp RepositoryRef {
-    /// Returns the Oid of `rev`.
-    pub fn rev_oid(&self, rev: &Rev) -> Result<Oid, Error>;
+    pub fn history(&self, rev: &Rev) -> Result<History, Error>;
 }
 ```
+TBD: how to retrieve the history of a particular file? Is it a valid use case?
```
FintanH, line 172: That should be possible. Basically, you want a specific revision and the file path to the blob. You can then get the commit, the tree that the commit points to, and find the blob via the path.

FintanH (commented):
I'm liking the additions :)

radicle-surf/docs/refactor-design.md
cloudhead: Yeah this is a super important use case! Histories of trees as well.

keepsimple1: I've updated the `History` API part (and doc) to support this.

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -183,10 +235,8 @@ impl<'a> RepositoryRef<'a> {
     /// Returns the last commit, if exists, for a `path` in the history of
     /// `rev`.
     pub fn last_commit(&self, path: file_system::Path, rev: &Rev) -> Result<Option<Commit>, Error> {
-        let git2_commit = self.rev_to_commit(rev)?;
-        let commit = Commit::try_from(git2_commit)?;
-        let file_history = self.file_history(&path, CommitHistory::Last, commit)?;
-        Ok(file_history.first().cloned())
+        let history = self.history(rev)?;
+        Ok(history.has_file(path).next())
```
keepsimple1, line 239: This shows how to use `has_file()` to get the history of a file.

radicle-surf/docs/refactor-design.md
```diff
@@ -143,18 +145,80 @@ The core API is:
...
 ```Rust
 imp RepositoryRef {
+    /// Returns the diff between two revisions.
     pub fn diff(&self, from: &Rev, to: &Rev) -> Result<Diff, Error>;
+
+    /// Returns the diff between a revision and the initial state of the repo.
+    pub fn initial_diff(&self, rev: &Rev) -> Result<Diff, Error> {
+
+    /// Returns the diff of a specific revision.
+    pub fn diff_of_rev(&self, rev: &Rev) -> Result<Diff, Error>;
```
keepsimple1, line 155: Is this useful at all?

radicle-surf/docs/refactor-design.md
```diff
@@ -143,18 +145,80 @@ The core API is:
...
+    /// Returns an iterator that only iterates commits that has the file `path`.
+    pub fn has_file(self, path: file_system::Path) -> impl Iterator<Item = Commit> + 'a;
+
+    /// Reloads the history so that a caller can iterate it again.
+    pub fn reload(&mut self);
+}
+```
+Note: the only way to create a new `History` to call `repo.history(rev)`.
+
+- Alternative design:
```
keepsimple1, line 192: What do you think of this alternative design?

radicle-surf/docs/refactor-design.md
```diff
@@ -143,18 +143,80 @@ The core API is:
...
+    /// This method creats a new `RevWalk` internally and return an
+    /// iterator for all commits in a history.
+    pub fn iter(&self) -> impl Iterator<Item = Commit>;
+}
+```
+
+In this design, `History` does not keep `RevWalk` in its state. It will create
+a new one when `iter()` is called. I like the immutable interface of this design
+but did not implement it in the current code mainly because the libgit2 doc says
+[creating a new `RevWalk` is relatively expensive](https://libgit2.org/libgit2/#HEAD/group/revwalk/git_revwalk_new).
```
keepsimple1, line 212: Here is a clickable link for the [libgit2 API doc](https://libgit2.org/libgit2/#HEAD/group/revwalk/git_revwalk_new) . If people are interested in this alternative design, I can do some measurements to see how "expensive" git_revwalk_new actually is.

radicle-surf/docs/refactor-design.md
FintanH: Remind me, is this the one that diffs the `rev` against its immediate parent? I think that case is useful because it's what's displayed if you navigate to a single commit.

radicle-surf/docs/refactor-design.md
FintanH: That's a super interesting find! Not only is it expensive it's also not thread-safe to share :scream:

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -35,23 +34,74 @@ use crate::{
...
+        &self.head
+    }
+
+    /// Returns an iterator that only iterates commits for a file specified by
+    /// `path`, in this history.
+    pub fn has_file(self, path: file_system::Path) -> impl Iterator<Item = Commit> + 'a {
+        let repo = self.repo.clone();
+        self.filter_map(
+            move |commit| match repo.repo_ref.find_commit(commit.id.into()) {
+                Ok(git2_commit) => match repo.diff_commit_and_parents(&path, &git2_commit) {
```
FintanH, line 66: This won't work with binary files due to [this line](https://github.com/radicle-dev/radicle-git/blob/be0f3aa06e580f4b6d1af6bbd12f7f68dc4a7153/radicle-surf/src/vcs/git/repo.rs#L433-L434), unfortunately.

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -35,23 +34,74 @@ use crate::{
...
+
+impl<'a> History<'a> {
+    /// Returns the first commit (i.e. the head) in the history.
+    pub fn head(&self) -> &Commit {
+        &self.head
+    }
+
+    /// Returns an iterator that only iterates commits for a file specified by
+    /// `path`, in this history.
+    pub fn has_file(self, path: file_system::Path) -> impl Iterator<Item = Commit> + 'a {
```
FintanH, line 62: Perhaps this should be `Result<Commit, Error>` so that errors aren't filtered out, but instead returned.

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -35,23 +34,74 @@ use crate::{
...
+                    Err(_) => None,
+                },
+                Err(_) => None,
+            },
+            None => None,
+        }
+    }
+}
+
+impl<'a> PartialEq for History<'a> {
```
FintanH, line 94: Is this needed?

radicle-surf/docs/refactor-design.md
keepsimple1: Yes, you're correct.

radicle-surf/src/vcs/git/repo.rs
FintanH: I see, I wonder if it still makes sense if it's only comparing the `head`. Before it would ensure that the whole history matched, i.e. the `NonEmpty<Commit>`.

I would either change the tests to compare the heads or have a compare that zips the history iterators and compares. I would maybe opt for the first since this code is only for tests. I don't think the user-facing API needs to compare histories.

radicle-surf/src/vcs/git/repo.rs
keepsimple1: I wanted to have the `Item` type consistent as the `History` iterator item type, i.e.` Commit`. I agree that this is a trade-off. If we do want to bubble up the potential errors, I think we should change `History` iterator returning `Result<>` as well.

radicle-surf/docs/refactor-design.md
keepsimple1: Just to add, the "not thread-safe" part would make the alternative design a bit more attractive.

radicle-surf/src/vcs/git/repo.rs
keepsimple1: Looks like this only affect a flag, but the binary file is still found.  I am adding a test case for this.

radicle-surf/src/vcs/git/repo.rs
keepsimple1: Changed the tests and removed this.

radicle-surf/t/src/git.rs
```diff
@@ -373,12 +351,23 @@ mod last_commit {
...
+            .history(&Branch::local("master").into())
             .unwrap()
-            .first()
+            .head()
             .id;
         assert_eq!(root_last_commit_id, Some(expected_oid));
     }
+
+    #[test]
+    fn binary_file() {
```
keepsimple1, line 362: Added this to check if we can find binary files in a history.  Also opened an issue https://github.com/radicle-dev/radicle-git/issues/40 that will look into the binary flag as well.

radicle-surf/src/vcs/git/commit.rs
```diff
@@ -166,3 +166,39 @@ impl<'repo> TryFrom<git2::Commit<'repo>> for Commit {
...
     }
 }
+
+/// Supports various ways to specify a commit in Git.
+pub trait IntoCommit {
+    /// Resolves into a commit in `repo`.
+    fn peel(&self, repo: &RepositoryRef) -> Result<Commit, Error>;
+}
+
+impl IntoCommit for &Commit {
```
FintanH, line 176: I'd make the `clone` happen by the caller, not the trait.
```suggestion
impl IntoCommit for Commit {
```

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -142,69 +190,69 @@ impl<'a> RepositoryRef<'a> {
...
-        }
-        .find_ref(self)?;
-        self.to_history(&reference)
-    }
-
     /// Get the [`Diff`] between two commits.
-    pub fn diff(&self, from: &Rev, to: &Rev) -> Result<Diff, Error> {
-        let from_commit = self.rev_to_commit(from)?;
-        let to_commit = self.rev_to_commit(to)?;
+    pub fn diff<R: FromRevision>(&self, from: R, to: R) -> Result<Diff, Error> {
```
FintanH, line 194: We're turning these into `Commit`s so this should use the `IntoCommit` trait :) The phrase "statement of intent" comes to mind. Since revisions can be used to refer to any objects they're too general, but `IntoCommit` tells us that we're definitely going through `Commit` objects.
```suggestion
    pub fn diff<R: IntoCommit>(&self, from: R, to: R) -> Result<Diff, Error> {
```

radicle-surf/src/vcs/git.rs
```diff
@@ -138,3 +141,35 @@ where
         })
     }
 }
+
+/// Supports various ways to specify a revision used in Git.
+pub trait FromRevision {
+    /// Resolves a revision into an object id in `repo`.
+    fn object_id(&self, repo: &RepositoryRef) -> Result<Oid, Error>;
```
FintanH, line 148: Not necessary but I'd probably call this `id`. There's no other types of ids lying around :)

radicle-surf/src/vcs/git.rs
```diff
@@ -138,3 +141,35 @@ where
...
+impl FromRevision for &str {
+    fn object_id(&self, _repo: &RepositoryRef) -> Result<Oid, Error> {
+        Oid::from_str(*self).map_err(Error::Git)
+    }
+}
+
+impl FromRevision for &Branch {
+    fn object_id(&self, repo: &RepositoryRef) -> Result<Oid, Error> {
+        let git2_oid = repo.repo_ref.refname_to_id(&self.refname())?;
+        Ok(git2_oid.into())
```
FintanH, line 166:
```suggestion
        Ok(repo.repo_ref.refname_to_id(&self.refname()).map(Oid::from)?)
```

radicle-surf/src/vcs/git.rs
```diff
@@ -138,3 +141,35 @@ where
...
+    }
+}
+
+impl FromRevision for &Tag {
+    fn object_id(&self, repo: &RepositoryRef) -> Result<Oid, Error> {
+        let git2_oid = repo.repo_ref.refname_to_id(&self.refname())?;
+        Ok(git2_oid.into())
```
FintanH, line 173:
```suggestion
        Ok(repo.repo_ref.refname_to_id(&self.refname()).map(Oid::from)?)
```

radicle-surf/src/vcs/git/commit.rs
```diff
@@ -166,3 +166,39 @@ impl<'repo> TryFrom<git2::Commit<'repo>> for Commit {
         })
     }
 }
+
+/// Supports various ways to specify a commit in Git.
+pub trait IntoCommit {
+    /// Resolves into a commit in `repo`.
+    fn peel(&self, repo: &RepositoryRef) -> Result<Commit, Error>;
```
FintanH, line 173: I'm just realising that this isn't a `git2::Commit`. I think for getting the most out of this trait it should return `git2::Commit`.

radicle-surf/src/vcs/git/tag.rs
```diff
@@ -106,6 +106,11 @@ impl Tag {
             Self::Annotated { name, .. } => name.clone(),
         }
     }
+
+    /// Returns the full ref name of the tag.
+    pub fn refname(&self) -> String {
+        format!("refs/tags/{}", self.name().name())
```
FintanH, line 112: At some point we should start using the `git-ref-format` crate

radicle-surf/docs/refactor-design.md
```diff
@@ -143,18 +143,80 @@ The core API is:
 
 ```Rust
 imp RepositoryRef {
```
FintanH, line 145: I noticed that you have a few `imp` typos which should be `impl` FYI :)

radicle-surf/docs/refactor-design.md
FintanH: Well, fun fact! The repository can't be shared across threads either :upside_down_face: In Rust this is indicated by it either having a `Sync` impl or not which in this case it's `!Sync` https://docs.rs/git2/0.15.0/git2/struct.Repository.html#impl-Sync-for-Repository.

Having the `Revwalk` inside means that we don't need to create a new one every time, but rather can reconfigure it for each walk. It also seems to [reset](https://docs.rs/git2/0.15.0/git2/struct.Revwalk.html#method.reset) after iteration completes :thinking: It might be worth playing around with scenarios of reusing the same `History` and see what happens.

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -35,23 +35,70 @@ use crate::{
...
+
+impl<'a> History<'a> {
+    /// Returns the first commit (i.e. the head) in the history.
+    pub fn head(&self) -> &Commit {
+        &self.head
+    }
+
+    /// Returns an iterator that only iterates commits for a file specified by
+    /// `path`, in this history.
+    pub fn has_file(self, path: file_system::Path) -> impl Iterator<Item = Commit> + 'a {
```
FintanH, line 63: Out of curiosity, is there a reason you didn't want to follow suit on a similar formulation of my sketch here https://github.com/radicle-dev/radicle-git/pull/39/files#diff-72117ea7d2958719f587571649f74bbdbdac6c8a54cad6d3ca0247189a85aefcR5? Looking to understand what you like or don't like about the approach :)

radicle-surf/docs/refactor-design.md
cloudhead: Yeah, I would call it something more specific though, they are all diffs of rev.. maybe `parent_diff` or `diff_from_parent`

radicle-surf/docs/refactor-design.md
```diff
@@ -143,18 +143,80 @@ The core API is:
...
+    head: Commit,
+    revwalk: git2::Revwalk<'a>,
+}
+
+impl<'a> History<'a> {
+    /// `head` is always valid. A history is guaranteed not emnpty.
+    pub fn head(&self) -> &Commit;
+
+    /// Returns an iterator that only iterates commits that has the file `path`.
+    pub fn has_file(self, path: file_system::Path) -> impl Iterator<Item = Commit> + 'a;
```
cloudhead, line 182: Hmm I would call this `by_path` or `filter_by_path`.

radicle-surf/docs/refactor-design.md
```diff
@@ -143,18 +143,80 @@ The core API is:
...
+
+    /// Reloads the history so that a caller can iterate it again.
+    pub fn reload(&mut self);
```
cloudhead, line 185: This is not typical of rust iterators, normally they are consumed by the iteration. I'm not sure why we would need to re-iterate.

radicle-surf/src/vcs/git.rs
```diff
@@ -138,3 +141,35 @@ where
...
 }
+
+/// Supports various ways to specify a revision used in Git.
+pub trait FromRevision {
+    /// Resolves a revision into an object id in `repo`.
+    fn object_id(&self, repo: &RepositoryRef) -> Result<Oid, Error>;
+}
+
+impl FromRevision for Oid {
+    fn object_id(&self, _repo: &RepositoryRef) -> Result<Oid, Error> {
```
cloudhead, line 152: Maybe I'm missing something, but why is the trait `FromRevision` if it's converting from a `RepositoryRef`? Shouldn't it be called `FromRepositoryRef`?

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -66,6 +113,7 @@ pub struct Repository(pub(super) git2::Repository);
 ///
 /// Use the `From<&'a git2::Repository>` implementation to construct a
 /// `RepositoryRef`.
+#[derive(Clone)]
```
cloudhead, line 116: `Copy` would also make sense for a reference type.

radicle-surf/src/vcs/git.rs
```diff
@@ -138,3 +141,35 @@ where
         })
     }
 }
+
+/// Supports various ways to specify a revision used in Git.
+pub trait FromRevision {
```
cloudhead, line 146: Seems like this should be called `ToOid`, based on the method.

radicle-surf/src/vcs/git/commit.rs
```diff
@@ -166,3 +166,39 @@ impl<'repo> TryFrom<git2::Commit<'repo>> for Commit {
         })
     }
 }
+
+/// Supports various ways to specify a commit in Git.
+pub trait IntoCommit {
```
cloudhead, line 171: Based on what this trait does, it would make more sense to call it `ToCommit`.

See: https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -142,69 +190,69 @@ impl<'a> RepositoryRef<'a> {
...
-        let commits = history.len();
+        let history = self.history(rev)?;
+        let mut commits = 0;
 
         let contributors = history
-            .iter()
-            .cloned()
-            .map(|commit| (commit.author.name, commit.author.email))
+            .map(|commit| {
+                commits += 1;
```
cloudhead, line 253: This is not necessary, as `BTreeSet` and `BTreeMap` keep their length internally.

See: https://doc.rust-lang.org/src/alloc/collections/btree/map.rs.html#178

radicle-surf/docs/refactor-design.md
keepsimple1: `fmt::Debug` for `History` might be able to use `reset` and push `head` again.

radicle-surf/src/vcs/git.rs
keepsimple1: I do feel `object_id` has more advantages than `id` in this case:  ;-)

`object_id` in Git is well defined, but `id` is less so. And the return type is `Oid`, not `Id`.

radicle-surf/src/vcs/git/repo.rs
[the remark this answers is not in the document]
keepsimple1: If the file history starts with a list of commits, and only when the user clicks / expands a commit we will show the actual file, then I think we can do the lazy loading. 

If we preload `File` as well for every commit, in addition to the extra memory, the history loading time can potentially be long for two reasons:  1) a file can be big 2) the history can be a long one with many commits.

radicle-surf/src/vcs/git/repo.rs
keepsimple1: In this case, `commits` is the number of commits, which is different (i.e. >=) from `BTreeSet` length, which is `contributors.len()` .

radicle-surf/src/vcs/git/commit.rs
keepsimple1: I can see that for some types like `&str`, `ToCommit` makes more sense.  For some other types like `Ref`, `IntoCommit` feels more natural.

If we name it as `ToCommit`, then maybe the method can be renamed to `to_commit`.

radicle-surf/src/vcs/git.rs
keepsimple1: Before this update, I was debating between `FromRevision` and `ToOid` (or `IntoOid`). Eventually I like the fact that `FromRevision` has "Revision" in it and "Revision" is an official [term in git](https://www.git-scm.com/docs/gitrevisions), which is what we like to support via this trait.

`ToOid` does not (explicitly) tell what kind of types we like to support, even though we could explain them in doc comments.

radicle-surf/docs/refactor-design.md
keepsimple1: I was using it when working on the patch. I am not sure if our users will need it. I can remove it for now.

radicle-surf/src/vcs/git/repo.rs
keepsimple1: IMO, I like the optional `FilterBy` and I will try to see if can pass it in a modify function to a `History`. I didn't like `BlobAt` as I don't think `History` needs to yield an actual `File` (or `Blob`) object. I also think maybe we can modify a `History` to get a history for a file, instead of a constructor `History::file()`.

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -35,23 +35,70 @@ use crate::{
...
+    }
 }
 
-/// A `History` that uses `git2::Commit` as the underlying artifact.
-pub type History = vcs::History<Commit>;
+impl<'a> Iterator for History<'a> {
+    type Item = Commit;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // If we encounter any errors, will return None.
```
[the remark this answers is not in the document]
keepsimple1: I will work on this.

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -35,23 +35,70 @@ use crate::{
...
-/// This is for flagging to the `file_history` function that it should
-/// stop at the first (i.e. Last) commit it finds for a file.
-pub(super) enum CommitHistory {
-    _Full,
-    Last,
+/// An iterator that produces the history of commits for a given `head`,
+/// in the `repo`.
+///
+/// As we have the `head`, this history is guaranteed not empty.
+pub struct History<'a> {
```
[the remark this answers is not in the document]
keepsimple1: Will work on this.

radicle-surf/docs/refactor-design.md
FintanH: > `fmt::Debug` for `History` might be able to use `reset` and push `head` again.

I don't understand, can you elaborate :)

radicle-surf/src/vcs/git/repo.rs
FintanH: Agreed, the iterator only returned `Commit` before it was safe to do so. Don't be afraid to change the code in these sort of cases as semantics change :)

radicle-surf/src/vcs/git/repo.rs
FintanH: Ah that's a fair point! I missed this function [`diff_blobs`](https://docs.rs/git2/0.15.0/git2/struct.Repository.html#method.diff_blobs).

> Are we sure this diff will only support commits going forward? (I know it's hard to predict the future, but I was thinking maybe it's beneficial to use `FromRevision` here).

I'm happy to go forward with `FromRevision`! Good catch :)

radicle-surf/src/vcs/git.rs
FintanH: I concede :bow:

radicle-surf/src/vcs/git/repo.rs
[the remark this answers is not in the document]
FintanH: > If we preload `File` as well for every commit, in addition to the extra memory, the history loading time can potentially be long for two reasons: 1) a file can be big 2) the history can be a long one with many commits.

Well since iterators are lazy, I don't think either of these points is a problem -- unless they get collected into a structure.

> If the file history starts with a list of commits, and only when the user clicks / expands a commit we will show the actual file, then I think we can do the lazy loading.

Based on [this](https://github.com/radicle-dev/radicle-git/commits/main/radicle-surf/src/lib.rs) though, it just gives a list of commits. Annoyingly though, if you click into the commit it gives you all the files, which sucks if you're just focusing on a single file.

FintanH: > IMO, I like the optional FilterBy and I will try to see if can pass it in a modify function to a History.

Ha! The `FilterBy` was what I didn't like about that approach :smile: It felt like a code smell waiting to happen as more filters would be added and wouldn't be used.

> I didn't like BlobAt as I don't think History needs to yield an actual File (or Blob) object.

Understood. What about returning the `Oid` so that the object can be retrieved easily?

radicle-surf/src/vcs/git.rs
FintanH: Ya, just to clarify a bit more -- I formalised this here https://github.com/radicle-dev/radicle-git/pull/38/files#diff-8fa8f2eb0bba44b7eb0f50532935cf0db8569f5441672e5acb7d9d91401887baR82-R109.

The idea is to resolve a git revision into its corresponding `Oid`.

radicle-surf/src/vcs/git/tag.rs
keepsimple1: Will look into this when starting on https://github.com/radicle-dev/radicle-git/issues/15 .

radicle-surf/docs/refactor-design.md
keepsimple1: Never mind, I was thinking that we could iterator the history to print out the commit list in `fmt::Debug` but then realized it does not take `&mut self`.

radicle-surf/src/vcs/git/repo.rs
keepsimple1: I've changed this method to return `History` as well.

radicle-surf/src/vcs/git/commit.rs
keepsimple1: thanks.

radicle-surf/src/vcs/git/commit.rs
keepsimple1: One of goals I listed is to [hide `git2` ](https://github.com/radicle-dev/radicle-git/blame/main/radicle-surf/docs/refactor-design.md#L27)in our new API.  I hope we can reach that goal.

radicle-surf/docs/refactor-design.md
keepsimple1: thanks!

radicle-surf/src/vcs/git/repo.rs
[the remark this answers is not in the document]
keepsimple1: done.

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -35,23 +35,70 @@ use crate::{
...
+        &self.head
+    }
+
+    /// Returns an iterator that only iterates commits for a file specified by
+    /// `path`, in this history.
+    pub fn has_file(self, path: file_system::Path) -> impl Iterator<Item = Commit> + 'a {
+        let repo = self.repo.clone();
+        self.filter_map(
+            move |commit| match repo.repo_ref.find_commit(commit.id.into()) {
+                Ok(git2_commit) => match repo.diff_commit_and_parents(&path, &git2_commit) {
```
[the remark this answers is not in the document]
keepsimple1: My understanding is that we wanted to list commits for the file only if there is a change to the file.  If that is not the case, then we can do it differently here.

radicle-surf/src/vcs/git/repo.rs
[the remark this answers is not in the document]
keepsimple1: Fixed to return a `Result` in `Item`.

radicle-surf/src/vcs/git/repo.rs
keepsimple1: I've updated to use `FilterBy` internally but not ask for them in `new()`. 

> Understood. What about returning the Oid so that the object can be retrieved easily?

I didn't get chance to work on this yet. I will play around with `by_path()` method again.

radicle-surf/docs/refactor-design.md
keepsimple1: updated.

radicle-surf/src/commit.rs
```diff
@@ -241,7 +243,13 @@ where
     };
 
     let stats = repo.get_stats(&rev)?;
-    let headers = repo.history(rev)?.iter().map(Header::from).collect();
+    let headers = repo
+        .history(&rev)?
+        .filter_map(|c| match c {
+            Ok(c) => Some(Header::from(c)),
+            Err(_) => None,
```
FintanH, line 250: Why choose to not return these errors?

radicle-surf/src/vcs/git.rs
```diff
@@ -138,3 +144,54 @@ where
...
```
FintanH, line 149: I'd say we should just call this `Revision` to avoid any confusion around the `From`

FintanH (commented):
Generally looks good to me :)

If you don't mind, I'll take a pass at the document this week to consolidate it with the design document I wrote. So the final product will give a high-level view as well as the implementation details you're providing in the redesign document.

radicle-surf/src/commit.rs
keepsimple1: Good point. I just learned that `Collect` can [do some magic with `Result`](https://doc.rust-lang.org/rust-by-example/error/iter_result.html#fail-the-entire-operation-with-collect). Will update.

radicle-surf/src/vcs/git.rs
keepsimple1: Sounds good to me.

radicle-surf/src/vcs/git/repo.rs
keepsimple1: I think we are good with this as `by_path` will return a `History` of commits for this file.  When iterating those commits, if the user wanted to open the file at that point, we can retrieve the object on-demand using the commit's Oid.  I.e. we don't have to pre-load and store the file's Oid earlier. This is similar with what GitHub does for a file history:  

Screenshot_20221031_102413

And the icon of opening the file at one point also links to a path under the commit Oid: 
Screenshot_20221031_102621

link: https://github.com/radicle-dev/radicle-git/blob/c3d9a2bdc511dbf9a78b6c5fc1b3ac6640595e56/radicle-surf/src/file_system/directory.rs

radicle-surf/src/vcs/git.rs
keepsimple1: Changed. In the next PR, I hope I can remove the other `Rev` type, otherwise it can also be confusing.

radicle-surf/src/commit.rs
keepsimple1: Fixed.

radicle-surf/src/vcs/git/repo.rs
keepsimple1: I added `get_file()` method for `Commit`, with a test case showing its usage. I think there is still room to iterate it. One thing is: should we add a repo reference in `Commit` so that its method calls can be more concise? I am leave it as is for this PR.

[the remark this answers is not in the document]
keepsimple1: As mentioned in the next comment, if you click the icon next to the commit link, then it will give you only the single file.

FintanH: > One thing is: should we add a repo reference in Commit so that its method calls can be more concise? I am leave it as is for this PR.

Mmmm that sounds like something I'd like to avoid. Carrying around the repo hidden in these objects comes across as a smell to me. You're hiding away the state rather than explicitly showing what functions do.

radicle-surf/docs/refactor-design.md
```diff
@@ -135,26 +215,142 @@ pub enum DirectoryContents {
...
+    pub fn iter(&self) -> impl Iterator<Item = Commit>;
+}
+```
+
+In this design, `History` does not keep `RevWalk` in its state. It will create
+a new one when `iter()` is called. I like the immutable interface of this design
+but did not implement it in the current code mainly because the libgit2 doc says
+[creating a new `RevWalk` is relatively expensive](https://libgit2.org/libgit2/#HEAD/group/revwalk/git_revwalk_new).
+
+### List refs and retrieve their metadata
```
FintanH, line 300: I'm not quite sure about the below but we can follow up. I think it'd be better to think about these once you're more familiar with `git-ref-format`.

PR #38, opened 2022-10-24
radicle-surf: design document

FintanH:
The work on the refactoring document[[0]] has been extremely useful in pinpointing problem areas in the radicle-surf redesign.

This document intends to supplement that by outlining a high-level API that the redesign can follow. The design specifies a series of components and recommended ways of implementing those components. The exact implementation details are avoided. It also aims to provide more education around different concepts in git by supplying git documentation and explanations of the different types one may come across in git.

[0]: https://github.com/radicle-dev/radicle-git/blob/main/radicle-surf/docs/refactor-design.md

### DIFF
--- a/radicle-surf/docs/design.md
+++ b/radicle-surf/docs/design.md
@@ -83,7 +83,7 @@ of indirection defined by `Vcs` trait.
 
 ## Components
 
-The `radicle-surf` library can split into a few main components for
+The `radicle-surf` library can be split into a few main components for
 browsing a `git` repository, each of which will be discussed in the
 following subsections.
 
@@ -91,6 +91,67 @@ Note that any of the API functions defined are _sketches_ and may not
 be the final form or signature of the functions themselves. The traits
 defined are recommendations, but other solutions for these
 representations may be discovered during implementation of this design.
+
+### Revisions
+
+Before describing the next components, it is important to first
+describe [revisions][git-revisions]. A revision in `git` is a way of
+specifying an `Oid`. This can be done in a multitude of ways. One can
+also specify a range of `Oid`s (think `git log`). The API will support
+taking revisions as parameters where an `Oid` is expected. It will
+not, however, permit ranges (at least for the time being) and so a
+revision will be scoped to any string that can resolve to a single
+`Oid`, e.g. an `Oid` string itself, a reference name, `@{date}`, etc.
+The aim will be to have a trait similar to:
+
+
+```rust
+/// `Self` is expected to be a type that can resolve to a single
+/// `Oid`.
+///
+/// An `Oid` is the trivial case and returns itself, and is
+/// infallible.
+///
+/// However, some other revisions require parsing and/or looking at the
+/// storage, which may result in an `Error`.
+pub trait FromRevision {
+  type Error;
+
+  /// Resolve the revision to its `Oid`, if possible.
+  fn peel(&self, storage: &Storage) -> Result<Oid, Self::Error>;
+}
+```
+
+#### Commit-ish
+
+The snapshot mentioned above is a `Commit` in `git`, where the
+commit points to a `Tree` object. Thus, the API should be able to take
+any parameter that may resolve to a `Commit`. This idea can be
+captured as a trait, similar to `FromRevision`, which allows something
+to be peeled to a `Commit`.
+
+```rust
+/// `Self` is expected to be a type that can resolve to a single
+/// `Commit`.
+///
+/// A `Commit` is the trivial case and returns itself, and is
+/// infallible.
+///
+/// However, some other kinds of data require parsing and/or looking at the
+/// storage, which may result in an `Error`.
+///
+/// Common cases are:
+///
+///   * Reference that points to a commit `Oid`.
+///   * A `Tag` that has a `target` of `Commit`.
+///   * An `Oid` that is the identifier for a particular `Commit`.
+pub trait Commitish {
+  type Error;
+
+  /// Resolve the type to its `Commit`, if possible.
+  fn peel(&self, storage: &Storage) -> Result<Commit, Self::Error>;
+}
+```
 
 ### References
 
@@ -148,35 +209,6 @@ to avoid retrieving the actual object to limit the amount of data
 needed. The `Oid` is the minimal amount of information required to
 fetch the object itself.
 
-### Revisions
-
-Before describing the next components, it is important to first
-describe [revisions][git-revisions]. A revision in `git` is a way of
-specifying an `Oid`. This can be done in a multitude of ways. One can
-also specify a range of `Oid`s (think `git log`). The API will support
-taking revisions as parameters where an `Oid` is expected. It will
-not, however, permit ranges (at least for the time being) and so a
-revision will be scoped to any string that can resolve to a single
-`Oid`, e.g. an `Oid` string itself, a reference name, `@{date}`, etc.
-The aim will be to have a trait similar to:
-
-```rust
-/// `Self` is expected to be a type that can resolve to a single
-/// `Oid`.
-///
-/// An `Oid` is the trivial case and returns itself, and is
-/// infallible.
-///
-/// However, some other revisions require parsing and/or looking at the
-/// storage, which may result in an `Error`.
-pub trait FromRevision {
-  type Error;
-
-  /// Resolve the revision to its `Oid`, if possible.
-  fn peel(&self, storage: &Storage) -> Result<Oid, Self::Error>;
-}
-```
-
 ### Objects
 
 Within the `git` model, [references][#References] point to
@@ -213,36 +245,7 @@ time. Generally, this object would be a `Tree`, i.e. a directory of
 files. However, it may be that a particular file, i.e. `Blob`, can be
 viewed.
 
-#### Commit-ish
-
-The snapshot mentioned above is a `Commit` in `git`, where the
-commit points to a `Tree` object. Thus, the API should be able to take
-any parameter that may resolve to a `Commit`. This idea can be
-captured as a trait, similar to `FromRevision`, which allows something
-to be peeled to a `Commit`.
-
-```rust
-/// `Self` is expected to be a type that can resolve to a single
-/// `Commit`.
-///
-/// A `Commit` is the trivial case and returns itself, and is
-/// infallible.
-///
-/// However, some other kinds of data require parsing and/or looking at the
-/// storage, which may result in an `Error`.
-///
-/// Common cases are:
-///
-///   * Reference that points to a commit `Oid`.
-///   * A `Tag` that has a `target` of `Commit`.
-///   * An `Oid` that is the identifier for a particular `Commit`.
-pub trait Commitish {
-  type Error;
-
-  /// Resolve the type to its `Commit`, if possible.
-  fn peel(&self, storage: &Storage) -> Result<Commit, Self::Error>;
-}
-```
+### Directories anf Files
 
 This provides the building blocks for defining common cases of viewing
 files and directories given a `Commitish` type.
--- a/radicle-surf/docs/design.md
+++ b/radicle-surf/docs/design.md
@@ -1,0 +1,266 @@
+# radicle-surf
+
+## Overview
+
+The main goal for the `radicle-surf` is to provide an API for
+accessing a `git` repository and providing a code browsing
+experience. This experience can be likened to GitHub or GitLab's
+project browsing pages. It does not aim to be an UI layer, but rather
+provides the functionality for a UI layer to built on top of it. With
+that in mind, this document sets out to define the main components of
+`radicle-surf` and a high-level design of the API. Note that this is
+the second iteration of designing this library -- where the first can
+be found in the [denotational-design.md][denotational] document.
+
+## Components
+
+The `radicle-surf` library can split into a few main components for
+browsing a `git` repository, each of which will be discussed in the
+following subsections.
+
+Note that any of the API functions defined are _sketches_ and may not
+be the final form or signature of the functions themselves. The traits
+defined are recommendations, but other solutions for these
+representations may be discovered during implementation of this design.
+
+### References
+
+Many are familiar with `git` branches. They are the main point of
+interaction when working within a `git` repository. However, the more
+general concept of a branch is a
+[reference][git-references]. References are stored within the `git`
+repository under the `refs/` directory. Within this directory, `git`
+designates a few specific [namespaces][git-references]:
+
+* `refs/heads` -- local branches
+* `refs/remotes` -- remote branches
+* `refs/tags` -- tagged `git` objects
+* `refs/notes` -- attached notes to `git` references
+
+These namespaces are designated special within `git`'s tooling, such
+as the command line, however, there is nothing stopping one from
+defining their own namespace, e.g. `refs/rad`.
+
+As well as this, there is another way of separating `git` references
+by a namespace which is achieved via the [gitnamespaces] feature. When
+`GIT_NAMESPACE` or `--git-namespace` is set, the references are scoped
+by `refs/namespaces/<namespace>`, e.g. if `GIT_NAMESPACE=rad` set then the
+`refs/heads/main` branch would mean
+`refs/namespaces/rad/refs/heads/main`.
+
+With the above in mind, the following API functions are suggested:
+
+```rust
+/// Return a list of references based on the `pattern` string supplied, e.g. `refs/rad/*`
+pub fn references(storage: &Storage, pattern: PatterStr) -> Result<References, Error>;
+
+/// Return a list of branches based on the `pattern` string supplied, e.g. `refs/heads/features/*`
+pub fn branches(storage: &Storage, pattern: BranchPattern) -> Result<Branches, Error>;
+
+/// Return a list of remote branches based on the `pattern` string supplied, e.g. `refs/remotes/origin/features/*`
+pub fn remotes(storage: &Storage, pattern: RemotePattern) -> Result<Remotes, Error>;
+
+/// Return a list of tags based on the `pattern` string supplied, e.g. `refs/tags/releases/*`
+pub fn tags(storage: &Storage, pattern: TagPattern) -> Result<Tags, Error>;
+
+/// Return a list of notes based on the `pattern` string supplied, e.g. `refs/notes/blogs/*`
+pub fn notes(storage: &Storage) -> Result<Notes, Error>;
+```
+
+It may be considered to be able to set an optional `gitnamespace`
+within the storage, or ammend the pattern types to allow for scoping
+by the `gitnamespace`.
+
+The returned list will not be the objects themselves. Instead they
+will be the metadata for those objects, i.e. `Oid`, `name`, etc. For
+the retrieval of those objects see the section on
+[Objects][#Objects]. The reason for this choice is that an UI may want
+to avoid retrieving the actual object to limit the amount of data
+needed. The `Oid` is the minimal amount of information required to
+fetch the object itself.
+
+### Revisions
+
+Before describing the next components, it is important to first
+describe [revisions][git-revisions]. A revision in `git` is a way of
+specifying an `Oid`. This can be done in a multitude of ways. One can
+also specify a range of `Oid`s (think `git log`). The API will support
+taking revisions as parameters where an `Oid` is expected. It will
+not, however, permit ranges (at least for the time being) and so a
+revision will be scoped to any string that can resolve to a single
+`Oid`, e.g. an `Oid` string itself, a reference name, `@{date}`, etc.
+The aim will be to have a trait similar to:
+
+```rust
+/// `Self` is expected to be a type that can resolve to a single
+/// `Oid`.
+///
+/// An `Oid` is the trivial case and returns itself, and is
+/// infallible.
+///
+/// However, some other revisions require parsing and/or looking at the
+/// storage, which may result in an `Error`.
+pub trait FromRevision {
+  type Error;
+
+  /// Resolve the revision to its `Oid`, if possible.
+  fn peel(&self, storage: &Storage) -> Result<Oid, Self::Error>;
+}
+```
+
+### Objects
+
+Within the `git` model, [references][#References] point to
+[objects][git-objects]. The types of objects in `git` are: commits, tags (lightweight
+& annotated), notes, trees, and blobs.
+
+All of these objects can retrieved via their `Oid`. The API will
+supply functions to retrieve them all for completion's sake, however,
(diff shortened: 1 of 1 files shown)


radicle-surf/docs/design.md
```diff
@@ -0,0 +1,266 @@
...
+`refs/namespaces/rad/refs/heads/main`.
+
+With the above in mind, the following API functions are suggested:
+
+```rust
+/// Return a list of references based on the `pattern` string supplied, e.g. `refs/rad/*`
+pub fn references(storage: &Storage, pattern: PatterStr) -> Result<References, Error>;
+
+/// Return a list of branches based on the `pattern` string supplied, e.g. `refs/heads/features/*`
+pub fn branches(storage: &Storage, pattern: BranchPattern) -> Result<Branches, Error>;
```
keepsimple1, line 58: Do we need the `BranchPattern` parameter? How common would our users like to provide an actual pattern, vs. providing no patterns? 

If providing an actual pattern is not common, we could consider to support it via a separate filter function.

radicle-surf/docs/design.md
```diff
@@ -0,0 +1,266 @@
...
+
+/// Return a list of remote branches based on the `pattern` string supplied, e.g. `refs/remotes/origin/features/*`
+pub fn remotes(storage: &Storage, pattern: RemotePattern) -> Result<Remotes, Error>;
+
+/// Return a list of tags based on the `pattern` string supplied, e.g. `refs/tags/releases/*`
+pub fn tags(storage: &Storage, pattern: TagPattern) -> Result<Tags, Error>;
+
+/// Return a list of notes based on the `pattern` string supplied, e.g. `refs/notes/blogs/*`
+pub fn notes(storage: &Storage) -> Result<Notes, Error>;
```
keepsimple1, line 67: It is not clear to me what the result types of the above functions are. I.e. what `Branches`, `Remotes`, `Tags`, `Notes` would look like.

radicle-surf/docs/design.md
```diff
@@ -0,0 +1,266 @@
...
+needed. The `Oid` is the minimal amount of information required to
+fetch the object itself.
+
+### Revisions
+
+Before describing the next components, it is important to first
+describe [revisions][git-revisions]. A revision in `git` is a way of
+specifying an `Oid`. This can be done in a multitude of ways. One can
+also specify a range of `Oid`s (think `git log`). The API will support
+taking revisions as parameters where an `Oid` is expected. It will
```
keepsimple1, line 88: I am thinking whether `Revision` or `Oid` should be the basic identifier.  And, whether `FromRevision` or `IntoOid` should be the trait to define in this context.

Further more, in the refactor design, I was thinking if we could define `CommitId` (it's an Oid) and not expose `Oid` in the API.  The rationale is that `Commit` is a concept even high-level user would know & use, but `Object` is probably not.

radicle-surf/docs/design.md
```diff
@@ -0,0 +1,266 @@
...
+pub fn commit<R: FromRevision>(storage: &Storage, rev: R) -> Result<Commit, Error>;
+
+/// Get the tag found by `oid`.
+pub fn tag<R: FromRevision>(storage: &Storage, rev: R) -> Result<Tag, Error>;
+
+/// Get the blob found by `oid`.
+pub fn blob<R: FromRevision>(storage: &Storage, rev: R) -> Result<Blob, Error>;
+
+/// Get the tree found by `oid`.
+pub fn tree<R: FromRevision>(storage: &Storage, rev: R) -> Result<Tree, Error>;
```
keepsimple1, line 133: Although many engineers know about `blob` and `tree` objects, they are Git internals.  For a high-level lib like `radicle-surf`, I think we should not expose them in the API, at least not initially, unless requested by our users explicitly.

radicle-surf/docs/design.md
```diff
@@ -0,0 +1,266 @@
...
+  fn peel(&self, storage: &Storage) -> Result<Commit, Self::Error>;
+}
+```
+
+This provides the building blocks for defining common cases of viewing
+files and directories given a `Commitish` type.
+
+```rust
+/// Get the `Directory` found at `commit`.
+pub fn directory<C: Commitish, P: AsRef<Path>>(
```
keepsimple1, line 183: Related to a previous comment, I am pondering in the refactor design that if we can define a `CommitId` and `IntoCommitId` trait. Then `directory` will just take a `CommitId` (not a `Commit` itself).

keepsimple1 (commented):
This naturally creates some conflicts with the mentioned refactor design doc and created some confusion for me. I discussed my confusions off this PR. Meanwhile, I provided some comments inline.

radicle-surf/docs/design.md
```diff
@@ -0,0 +1,266 @@
...
+defining their own namespace, e.g. `refs/rad`.
+
+As well as this, there is another way of separating `git` references
+by a namespace which is achieved via the [gitnamespaces] feature. When
+`GIT_NAMESPACE` or `--git-namespace` is set, the references are scoped
+by `refs/namespaces/<namespace>`, e.g. if `GIT_NAMESPACE=rad` set then the
+`refs/heads/main` branch would mean
+`refs/namespaces/rad/refs/heads/main`.
+
+With the above in mind, the following API functions are suggested:
```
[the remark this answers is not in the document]
FintanH: Ya, good point! Maybe to reiterate -- and perhaps I didn't explain this well -- the functions are merely sketches and it's the functionality I wanted to focus on. I mention that [here](https://github.com/radicle-dev/radicle-git/pull/38/files#diff-8fa8f2eb0bba44b7eb0f50532935cf0db8569f5441672e5acb7d9d91401887baR21-R24).

But I can also mention that `Storage` is deliberately opaque and merely represents any way of accessing a `git` repository. To clarify, this attempt doesn't care about the backing storage -- it just cares about the functions (in whatever form) exist to use.

radicle-surf/docs/design.md
```diff
@@ -0,0 +1,266 @@
...
+
+```rust
+/// Return a list of references based on the `pattern` string supplied, e.g. `refs/rad/*`
+pub fn references(storage: &Storage, pattern: PatterStr) -> Result<References, Error>;
```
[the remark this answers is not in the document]
FintanH: > This general function overlaps with more specific functions below (e.g. branches) in the sense that the user can use either of them to get the local branches (or other types of references).

I deliberately added this as a general case! :) This is to leave the API open for upstream users to fetch references under namespaces that they define themselves. The `cobs` use case is a prime example -- the clients team may want to be able to fetch collaborative objects and they could use `radicle-surf` tooling for that. Would that be right, @cloudhead?

> Will `Reference` be an enum of different type of references? Or will it be basically a string? If it's an enum, we can get away without other functions for special reference types. On the other hand, special functions provide better type checking.

`References` would be an iterator over the references that match the `pattern`. Each reference would be their name and their target -- `Oid` for direct references and another reference for symbolic references.

The same would go for `Branches`, but the reason it has another type is that the names are _expected_ to be prefixed with `refs/heads`. Similar for `Tags`, etc.

radicle-surf/docs/design.md
FintanH: That's a good point! I guess we could go either way. If a `pattern` was taken then we could define a constant:
```rust
lazy_static! {
  static ref ALL: BranchPattern = pattern_str!("refs/heads/*");
}
```

As long as `Branches` stays an iterator, I think the filter approach after works equally as well! :)

radicle-surf/docs/design.md
FintanH: Ya, I probably should have been more clear but as mentioned here, https://github.com/radicle-dev/radicle-git/pull/38#discussion_r1004136237, they would be iterators over references but with new types to signal they start with their correct namespaces.

radicle-surf/docs/design.md
FintanH: > I am thinking whether `Revision` or `Oid` should be the basic identifier. And, whether `FromRevision` or `IntoOid` should be the trait to define in this context.

So the reason I was thinking `FromRevision` is useful is that it makes function calls more flexible. For example, say I want a specific `Commit`, I may know a reference name that points to it, or its full SHA1, or even it's short-SHA1. So, all these function calls would result in the same `Commit`:
```
commit(storage, refname!("refs/heads/surf-design"))
commit(storage, "e9e23f49641b6b9a850969e5fb87f301fc10d2b9")
commit(storage, "e9e23f49641b6b9")
```

Does that make sense?

> Further more, in the refactor design, I was thinking if we could define `CommitId` (it's an Oid) and not expose `Oid` in the API. The rationale is that `Commit` is a concept even high-level user would know & use, but `Object` is probably not.

This is what I was calling `Commitish` :) This is terminology that I've come across in `git` docs quite frequently and here's an SO answer that explains it https://stackoverflow.com/a/23303550

So, yes I think the high-level would be exposing a `Commitish` or something of that kind. Again, I define it as a trait to be more flexible. Allowing `Oid`s, reference names, or whatever can resolve to a `Commit`. Our main users, the rest of the clients team, are familiar with `Oid`s and `Object`s but we can hide some of those things too :)

radicle-surf/docs/design.md
FintanH: That's fair! I still think it would be good to have them available. They may be available through an `objects` module but not at the top level.

I think for libraries like this, it's better to have the functionality available. Just because we can't imagine the use case doesn't mean it doesn't exist yet :grin:

radicle-surf/docs/design.md
FintanH: As I touched on here, https://github.com/radicle-dev/radicle-git/pull/38/files#r1004154254, this _can_ be an `Oid` for a `Commit` :grin: Or anything else that resolves to a `Commit`, e.g. a reference, a reference name, a tag object that points to a commit, a tag name that resolves to a tag object that points to a commit, etc. :smile:

radicle-surf/docs/design.md
```diff
@@ -0,0 +1,266 @@
...
+/// An `Oid` is the trivial case and returns itself, and is
+/// infallible.
+///
+/// However, some other revisions require parsing and/or looking at the
+/// storage, which may result in an `Error`.
+pub trait FromRevision {
+  type Error;
+
+  /// Resolve the revision to its `Oid`, if possible.
+  fn peel(&self, storage: &Storage) -> Result<Oid, Self::Error>;
```
jorgeig-space, line 107: How would this be used? If I understand correctly, a revision takes a [multitude of forms](https://git-scm.com/docs/gitrevisions#_specifying_revisions).

So we would have a `Commit`, `Tree`, etc. struct and then run `the_commit.peel(&storage)`? But at that point, we cannot have the object yet (because we don't know which commit we are talking about). Would this make more sense as an associated function?

radicle-surf/docs/design.md
```diff
@@ -0,0 +1,266 @@
...
+pub struct Old<C>(C);
+
+/// New type to differentiate the new side of a [`diff`].
+pub struct New<C>(C);
+
+/// Get the difference between the `old` and the `new` directories.
+pub fn diff<C: Commitish>(old: Old<C>, new: New<C>) -> Result<Diff, Error>
+```
+
+## Conclusion
```
jorgeig-space, line 252: I think a missing point is a high-level overview of how this API would be used for some common tasks:
- Show all files and directories in a repository
- Show the repo history
- Show the history of a file
- Show a file's content for a specific commit (I assume this is part of the objectives of the API)

For example:

### Show all files and directories at root of main branch of repository:
```
let rev = Revision::from_str("refs/heads/main");
let root_commit = Commit::from_rev(rev);
let root_dir = directory(&storage, root_commit);

// process root_dir iterator

File: radicle-surf/docs/design.md (hunk 8 above)
Comment (FintanH) [reply radicle-surf/docs/design.md:107]: Good question. So below we have a `commit` function -- as well as others, but I think `commit` will be the most commonly used. It takes a `FromRevision`. So let's see some examples of types that `impl FromRevision`:

```rust
impl FromRevision for Oid { /* code */ }

/// &str can be parsed into an `Oid`.
impl FromRevision for &str { /* code */ }

/// A reference can be looked up in the repository and its `Oid` is returned
impl FromRevision for RefString { /* code */ }

/// [ref]@{date}
/// https://git-scm.com/docs/gitrevisions#Documentation/gitrevisions.txt-emem
pub struct RefDate { name: Option<RefString>, date: Date }

impl FromRevision for RefDate { /* code */ }

/// [ref]@{n}
/// https://git-scm.com/docs/gitrevisions#Documentation/gitrevisions.txt-emltrefnamegtltngtemegemmaster1em
pub struct RefOrdinal { name: Option<RefString>, ordinal: usize }

impl FromRevision for RefOrdinal { /* code */ }
```

Now we can use _any_ of these in `commit`:
```rust
commit(storage, refname!("refs/heads/main"))
commit(storage, "e9e23f49641b6b9a850969e5fb87f301fc10d2b9")
commit(storage, "main@{n}".parse::<RefOrdinal>().unwrap())
```

Is that clear? Would adding this to the document help?

radicle-surf/docs/design.md
FintanH: So, are you saying an examples section would be useful? :)

jorgeig-space: Yes, that's my long-winded way of saying that :D

Like "examples of how we think this API would be used". I can write that tomorrow as a way of giving this a second review.

radicle-surf/docs/design.md
```diff
@@ -0,0 +1,266 @@
...
+
+### History
+
+Since `Commit`s in `git` form a history of changes via a linked-list,
+i.e. commits may have parents, and those parents grand-parents etc.,
+it is important that an API for iterating through the history is
+provided.
+
+The general mechanism for looking at a history of commits is called a
+[`revwalk`][libgit-revwal] in most `git` libraries. This provides a
```
FintanH, line 208:
```suggestion
[`revwalk`][libgit-revwalk] in most `git` libraries. This provides a
```

radicle-surf/docs/design.md
FintanH: Ahaha, I just wanted to clarify :pray:

radicle-surf/docs/design.md
jorgeig-space: Ah, I understood `Self` in this case was a `Commit`, `Tree`, etc. that had an `Oid`. But the idea is to implement the trait for everything that may express an `Oid`, not for the entities that are identified by one.

Yeah, then it makes sense. I'd add those examples so that it's clear, without having to write prose about "representing an Oid instead of being an object that is represented by an Oid."

But from the examples, would it make sense to equate this trait to `TryInto`/`Into`? I don't have a strong opinion on having a specific trait vs. `TryInto`, so just leaving this here for thought since I think we're going to have a lot of traits flying around 😅 .

FintanH: > Ah, I understood `Self` in this case was a `Commit`, `Tree`, etc. that had an `Oid`. But the idea is to implement the trait for everything that may express an `Oid`, not for the entities that are identified by one.

Aye, hence my whole spiel about git revisions and linking to the documentation :smile: 

> Yeah, then it makes sense. I'd add those examples so that it's clear, without having to write prose about "representing an Oid instead of being an object that is represented by an Oid."

Makes sense, can do!

> But from the examples, would it make sense to equate this trait to `TryInto`/`Into`? I don't have a strong opinion on having a specific trait vs. `TryInto`, so just leaving this here for thought since I think we're going to have a lot of traits flying around 

While that _would_ be nice, all the types would have to be augmented with the backing store so that `Oid`s could be looked up. Having a new trait allows us to say that storage is a parameter.


With that being more clear now, what's your opinion on this formulation? Useful or not useful? Would you prefer/see a different approach?

jorgeig-space: I think this is definitely useful - I'll walk through a few scenarios (listing the repo, see the history of a file, etc.) to ensure we're not missing anything but I think this formulation makes sense.

radicle-surf/docs/design.md
keepsimple1: I think we can start with using `file`, `directory` in our API, and only support `blob`, `tree` when users request them in the future. (Unless users have already requested ;-) )

radicle-surf/docs/design.md
keepsimple1: I thought more about this, and I think your approach make sense. I've adopted your suggestions and used traits for the API changes in PR https://github.com/radicle-dev/radicle-git/pull/37 , with small changes to the naming. Thanks!

FintanH: > I thought more about this, and I think your approach make sense. I've adopted your suggestions and used traits for the API changes in PR #37 , with small changes to the naming. Thanks!

Amazing, teamwork makes the dreamwork :sunglasses:

radicle-surf/docs/design.md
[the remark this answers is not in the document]
cloudhead: Yeah, this is pretty standard and we wouldn't want to lose this when moving from git2.

radicle-surf/docs/design.md
```diff
@@ -0,0 +1,266 @@
...
+```rust
+/// `Self` is expected to be a type that can resolve to a single
+/// `Oid`.
+///
+/// An `Oid` is the trivial case and returns itself, and is
+/// infallible.
+///
+/// However, some other revisions require parsing and/or looking at the
+/// storage, which may result in an `Error`.
+pub trait FromRevision {
```
cloudhead, line 103: `From*` traits usually have a function that returns `Self`, but here we return an `Oid`. Wouldn't it make more sense to call this simply `Revision`? A `Revision` is any type that can be peeled to an `Oid`..

radicle-surf/docs/design.md
cloudhead: file/directory only makes sense for working copies; if this library is going to be used on bare repositories, I don't see how we can get around blob and tree?

radicle-surf/docs/design.md
```diff
@@ -0,0 +1,266 @@
...
+///
+///   * Reference that points to a commit `Oid`.
+///   * A `Tag` that has a `target` of `Commit`.
+///   * An `Oid` that is the identifier for a particular `Commit`.
+pub trait Commitish {
+  type Error;
+
+  /// Resolve the type to its `Commit`, if possible.
+  fn peel(&self, storage: &Storage) -> Result<Commit, Self::Error>;
+}
```
cloudhead, line 175: What's the difference between a revision and a commitish?

radicle-surf/docs/design.md
FintanH: Ya, I originally thought `FromRevision` because it's going _from_ a revision _to_ an `Oid`, but `Revision` works for me too

radicle-surf/docs/design.md
FintanH: Would they not be equivalently used, no? I've advised to have these functions available in a module and not expose them at the top-level module. 

This is another point for standalone functions @keepsimple1. You can hide them because they're not attached to the struct as methods :smile:

radicle-surf/docs/design.md
FintanH: The idea is that a revision can be used to refer to _any_ type of object whereas commitish implies that whatever you pass needs to resolve to a `Commit`. Technically, revision implies commitish so you could have:

```rust
/// The `Revision` *must* resolve to a `Commit`.
impl<R: Revision> Commitish for R {
  type Error = Error;
  
  fn peel(&self, storage: &Storage) -> Result<Commit, Self::Error> {
    let oid = Revision::peel(self)?;
    /* try resolve it to a Commit, i.e. find_commit, find_tag -> commit, etc.*/
  }
}
```

cloudhead: Hmm, I thought that's what a revision was. See https://git-scm.com/docs/git-rev-parse#_specifying_revisions

I think if you're trying to retrieve a blob for example, you wouldn't expect to pass a revision, but rather an oid directly, since named refs typically don't point to blobs, only commits.

radicle-surf/docs/design.md
cloudhead: Cool, yeah I think it would work better, because `FromRevision` implies that the implementor of the trait isn't a `Revision`, since it's being converted "from a revision"; but in this case, all trait implementors *are* valid revisions.

radicle-surf/docs/design.md
cloudhead: > Would they not be equivalently used, no? I've advised to have these functions available in a module and not expose them at the top-level module.

Yeah, I mean if it's just a question of naming, and File has all the capabilities of Blob, and Directory has all the capabilities of Tree, then it could work; although I'd say it's a little confusing because it's not an abstraction, just a rename.

radicle-surf/docs/design.md
FintanH: > A revision parameter <rev> typically, but not necessarily, names a commit object.

_but not necessarily_ :)

> I think if you're trying to retrieve a blob for example, you wouldn't expect to pass a revision, but rather an oid directly, since named refs typically don't point to blobs, only commits.

Ya, typically. But there's also revisions that point to tags :)

radicle-surf/docs/design.md
FintanH: Ya, that's fair. I would opt for a non-specialised `tree`/`blob` and then a specialised `file`/`directory` because they won't just be renamed.

radicle-surf/docs/design.md
cloudhead: > Ya, typically. But there's also revisions that point to tags :)

But tags resolve to commits in the end.. I guess I'd be curious to see examples when you'd accept a Commitish as a function argument, and accepting a Revision wouldn't work, or something like that, which would be a motivating example.

radicle-surf/docs/design.md
keepsimple1: > file/directory only makes sense for working copies; if this library is going to be used on bare repositories, I don't see how we can get around blob and tree?

GitHub repos are bare repos but they still present "files" and "directories".  (They might did some internal extra work for that, but it is an implementation detail). 

I guess that "files" and "directories" are more identified by "paths", while "blob" and "tree" are identified by their `Oid`.  We could support both. I just think if I were to use radicle-surf to write an UI I would just use `file` and `directory` as much as possible.

cloudhead: Yeah, I think it depends on whether we're ok with users of `surf` to still require direct access to the `git2` repository. If that's by design, then we could get away with File/Directory. But if the plan is for users of this library to not need `git2` access, then we might need access to these lower level objects.

keepsimple1: I would not let users of `surf` to have direct access to `git2`. I am ok if it turns out that we need to provide users access to lower-level objects via our API.

PR #39, opened 2022-10-24
WIP: radicle-surf history

FintanH:
Based on the design document at https://github.com/radicle-dev/radicle-git/pull/38, I decided to play around with the history formulation using a revwalk. It looks promising!

I wanted to see what the output would look like so I added a test that fails and run:
`cargo test -p radicle-surf-test history`

### DIFF
--- a/radicle-surf/src/history.rs
+++ b/radicle-surf/src/history.rs
@@ -1,0 +1,118 @@
+use std::{marker::PhantomData, path::PathBuf};
+
+use git2::Revwalk;
+
+pub struct History<'a, A> {
+    walk: Revwalk<'a>,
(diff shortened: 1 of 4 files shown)


keepsimple1 (commented):
Thank you for sharing your ideas and code. I like your idea of defining `History` as an iterator (or produces an iterator for that matter). I felt we are on the same page at the high level. I made changes in my PR https://github.com/radicle-dev/radicle-git/pull/37 with similar design, but also different in various aspects. I am OK with either moving forward using this PR or using my PR.

I will skip commenting on details here for now as the difference in my PR probably says things more clearly.  I've also included an alternative design in the updated doc that might be worth considering as well. Thanks!

PR #43, opened 2022-11-01
git-commit: introduce git-commit crate

FintanH:
Porting the commit code that is found in heartwood[[0]]. It represents a parsed git commit object -- sans the id for the commit itself.

[0]: https://github.com/radicle-dev/heartwood/blob/16b22d9178fa6a1aaf2ecb01ec97e067399c5286/radicle/src/git.rs#L392-L575

### DIFF
--- a/git-commit/Cargo.toml
+++ b/git-commit/Cargo.toml
@@ -16,3 +16,7 @@ thiserror = "1"
 version = "0.15.0"
 default-features = false
 features = ["vendored-libgit2"]
+
+[dependencies.git-trailers]
+version = "0.1.0"
+path = "../git-trailers"
--- a/git-commit/Cargo.toml
+++ b/git-commit/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "git-commit"
+version = "0.1.0"
+license = "MIT OR Apache-2.0"
+edition = "2021"
+authors = [
+  "Alexis Sellier <alexis@radicle.xyz>",
+  "Fintan Halpenny <fintan.halpenny@gmail.com>",
+]
+keywords = ["git", "radicle"]
+
+[dependencies]
+thiserror = "1"
+
+[dependencies.git2]
+version = "0.15.0"
+default-features = false
+features = ["vendored-libgit2"]
--- a/git-commit/src/headers.rs
+++ b/git-commit/src/headers.rs
@@ -1,0 +1,71 @@
+use std::borrow::Cow;
+
+const BEGIN_SSH: &str = "-----BEGIN SSH SIGNATURE-----\n";
+const BEGIN_PGP: &str = "-----BEGIN PGP SIGNATURE-----\n";
+
+/// A collection of headers stored in a [`super::Commit`].
+///
+/// Note: these do not include `tree`, `parent`, `author`, and `committer`.
+#[derive(Clone, Debug, Default)]
+pub struct Headers(pub(super) Vec<(String, String)>);
+
+/// A `gpgsig` signature stored in a [`super::Commit`].
+pub enum Signature<'a> {
+    /// A PGP signature, i.e. starts with `-----BEGIN PGP SIGNATURE-----`.
+    Pgp(Cow<'a, str>),
+    /// A SSH signature, i.e. starts with `-----BEGIN SSH SIGNATURE-----`.
+    Ssh(Cow<'a, str>),
+}
+
+impl<'a> Signature<'a> {
+    fn from_str(s: &'a str) -> Result<Self, UnknownScheme> {
+        if s.starts_with(BEGIN_SSH) {
+            Ok(Signature::Ssh(Cow::Borrowed(s)))
+        } else if s.starts_with(BEGIN_PGP) {
+            Ok(Signature::Pgp(Cow::Borrowed(s)))
+        } else {
+            Err(UnknownScheme)
+        }
+    }
+}
+
+pub struct UnknownScheme;
+
+impl<'a> ToString for Signature<'a> {
+    fn to_string(&self) -> String {
+        match self {
+            Signature::Pgp(pgp) => pgp.to_string(),
+            Signature::Ssh(ssh) => ssh.to_string(),
+        }
+    }
+}
+
+impl Headers {
+    pub fn new() -> Self {
+        Headers(Vec::new())
+    }
+
+    pub fn iter(&self) -> impl Iterator<Item = (&str, &str)> {
+        self.0.iter().map(|(k, v)| (k.as_str(), v.as_str()))
+    }
+
+    pub fn values<'a>(&'a self, name: &'a str) -> impl Iterator<Item = &'a str> + '_ {
+        self.iter()
+            .filter_map(move |(k, v)| (k == name).then_some(v))
+    }
+
+    pub fn signatures(&self) -> impl Iterator<Item = Signature> + '_ {
+        self.0.iter().filter_map(|(k, v)| {
+            if k == "gpgsig" {
+                Signature::from_str(v).ok()
+            } else {
+                None
+            }
+        })
+    }
+
+    /// Push a header to the end of the headers section.
+    pub fn push(&mut self, name: &str, value: &str) {
+        self.0.push((name.to_owned(), value.trim().to_owned()));
+    }
+}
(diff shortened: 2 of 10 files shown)


git-commit/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "git-commit"
+version = "0.1.0"
+license = "MIT OR Apache-2.0"
+edition = "2021"
+authors = [
+  "Alexis Sellier <alexis@radicle.xyz>",
+  "Fintan Halpenny <fintan.halpenny@gmail.com",
```
cloudhead, line 8:
```suggestion
  "Fintan Halpenny <fintan.halpenny@gmail.com>",
```
If you copy pasted this, I'd check the other tomls just in case xD

git-commit/src/headers.rs
```diff
@@ -0,0 +1,92 @@
...
+
+const BEGIN_SSH: &str = "-----BEGIN SSH SIGNATURE-----\n";
+const END_SSH: &str = "\n-----END SSH SIGNATURE-----";
+
+const BEGIN_PGP: &str = "-----BEGIN PGP SIGNATURE-----\n";
+const END_PGP: &str = "\n-----END PGP SIGNATURE-----";
+
+/// A collection of headers stored in a [`super::Commit`].
+///
+///
```
cloudhead, line 11:
```suggestion
```

git-commit/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+                    .0
+                    .last_mut()
+                    .map(|(_, v)| v)
+                    .ok_or(ParseError::InvalidFormat)?;
+                value.push('\n');
+                value.push_str(rest);
+                continue;
+            }
+
+            if let Some((name, value)) = line.split_once(' ') {
```
cloudhead, line 169: You should lead with this, then you can just `match (k, v) { ("commiter", value) => ..., (.., ..) => }`

git-commit/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+        self.tree
+    }
+
+    /// The parent [`Oid`]s of this commit.
+    pub fn parents(&self) -> impl Iterator<Item = Oid> + '_ {
+        self.parents.iter().copied()
+    }
+
+    /// The author of this commit, i.e. the header corresponding to `author`.
+    pub fn author(&self) -> &Person {
```
cloudhead, line 49: I'd return a `git2::Signature` here maybe, for compatibility, and to not introduce a new type.

https://docs.rs/git2/latest/git2/struct.Signature.html

git-commit/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+        let mut buf = String::new();
+
+        buf.push_str(&format!("tree {}\n", self.tree));
+
+        for parent in &self.parents {
+            buf.push_str(&format!("parent {}\n", parent));
+        }
+
+        buf.push_str(&format!("author {}\n", self.author.to_string()));
+        buf.push_str(&format!("committer {}\n", self.committer.to_string()));
```
cloudhead, line 197: maybe `writeln!(buf, "author {}")` is more concise after importing `std::fmt::Write`[0]

[0]: https://doc.rust-lang.org/std/fmt/trait.Write.html

git-commit/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+                })?
+                .ok_or(ParseError::Missing("tree"))?,
+            None => return Err(ParseError::Missing("tree")),
+        };
+
+        let mut parents = Vec::new();
+        let mut author: Option<Person> = None;
+        let mut committer: Option<Person> = None;
+        let mut headers = Headers::new();
+
```
cloudhead, line 136: I'm curious why you opted to parse some of these headers separately from the rest, ie. why not just have `Headers` and be done with it. The code is much more complicated for it, and I'm not sure what value we get? Is this meant to be used somewhere?

git-commit/src/person.rs
```diff
@@ -0,0 +1,71 @@
...
+        buf.push(' ');
+        buf.push_str(&format!(
+            "{} {}{:0>4}",
+            self.time.seconds(),
+            self.time.sign(),
+            self.time.offset_minutes().abs()
+        ));
+        buf
+    }
+}
```
cloudhead, line 27: I'd try to use https://docs.rs/git2/latest/src/git2/signature.rs.html#150-159 here

cloudhead: I see you have a TryFrom<git2::Signature>, but I'd argue we should maybe just use that directly?

git-commit/src/lib.rs
FintanH: Ya, good point :+1:

git-commit/src/person.rs
FintanH: A couple of things:
* `git2::Signature` has a lifetime -- which is kind of annoying, but perhaps it can be static in this case.
* I mentioned above it'd be nice to abstract away the `git2` -- admittedly I didn't do it with `Time`.

Happy to go with `git2::Signature` though

git-commit/Cargo.toml
FintanH: Ahahaha, no I'm pretty sure I typed it :see_no_evil:

git-commit/src/person.rs
FintanH: The `Display` for `Signature` ignores the time :upside_down_face: https://docs.rs/git2/latest/src/git2/signature.rs.html#150-159

git-commit/src/lib.rs
cloudhead: Ok gotcha, so like we would return this type instead of `git2::Commit`. In that case it makes sense but I'd make sure to test it against a bunch of commits/repo somehow :thinking:

git-commit/src/person.rs
cloudhead: Yeah, see my other comment :+1:

git-commit/src/lib.rs
FintanH: I've tested this by adding a `read` function and creating a copy of this repo and parsing the first 20 commits. Everything looks good :)

FintanH:
@cloudhead do you wanna give the fixups a look and see what you think :)

I tested the signature flow in radicle-cob and all the tests are passing so everything seems to be working :+1:

git-commit/src/person.rs
```diff
@@ -1,36 +1,89 @@
...
+
+impl From<Time> for git2::Time {
+    fn from(t: Time) -> Self {
+        Self::new(t.seconds, t.offset)
+    }
+}
+
+impl fmt::Display for Time {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{} {}{:0>4}", self.seconds, self.sign, self.offset.abs())
```
cloudhead, line 65: Why not print the relative offset here? Seems a bit over the top to have a `Sign` type.. At the very least (if for some reason the offset doesn't display how you want), we could check here if the offset is positive or negative, since it seems like `sign` is only used here.

git-commit/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+use std::{
+    fmt::Write as _,
+    str::{self, FromStr},
+};
+
+use git2::{ObjectType, Oid};
+
+pub mod person;
+pub use person::Person;
```
cloudhead, line 22: Nit: what if it's a bot or a no-reply address :thinking: -- can't think of a better name though, since `Signature` means other things... maybe `Author` works better even, in the committer field..

cloudhead: `Actor` is another option :shrug: -- since this is a public type I think it's good to get it right.

git-commit/src/person.rs
FintanH: > Why not print the relative offset here? Seems a bit over the top to have a `Sign` type.. At the very least (if for some reason the offset doesn't display how you want), we could check here if the offset is positive or negative, since it seems like `sign` is only used here.

Ya, the formatting doesn't do what I want it to do https://doc.rust-lang.org/std/fmt/index.html#sign0. It'll omit a `0` if it's negative. But you're right, I can clean up the `Sign` and just check the `u32`

git-commit/src/lib.rs
FintanH: I'd probably go with `Author` since `Actor` is a bit overloaded in the tech world. Another option is `User` because the details do tend to come from `user.name` and `user.email`.

cloudhead: Agreed, Author > Actor. Whatever it is, make sure it's the same as in cobs..

Issue #13, opened 2022-09-21
Add `RefScope` as argument to `git::repo::revision_branches`


sebastinez:
Currently we retrieve all branch references, iterate over all of them and check if the passed Oid is contained in them.
https://github.com/radicle-dev/radicle-git/blob/58d69ebf7f441cfde043593fc7efbc8912033759/radicle-surf/src/vcs/git/repo.rs#L259

I think we could optimize this, by adding a `RefScope` argument e.g.
`fn revision_branches(&self, oid: &Oid, scope: RefScope) -> Result<Vec<Branch>, Error>`

This would allow us to initially create a Vector and retrieve only the local, or only remote or both if needed, and then check over that vector if the Oid is contained in any branch.

FintanH:
So with the unification of all git things into this repo, I think the best route is actually doing away with these custom types and using the `git-ref-format` package. This operation would then be defined in terms of a `PatternStr`, something like:

```rust
pub fn revision_branches(&self, oid: &Oid, pat: Option<PatternStr>) -> Result<Vec<RefString>, Error> {
   let refs = self.references(pat)?;
   Ok(refs.filter_map(|r| if contains(r, oid) then Some(r) else None))
}
```

If waiting for this refactor is a problem though, then an `Option<RefScope>` works too :)

sebastinez:
Ok nice, yeah it's not something that cannot wait, but will eventually provide a PR just to make it nicer for the `http-api` in the meanwhile 😁

FintanH:
> Ok nice, yeah it's not something that cannot wait, but will eventually provide a PR just to make it nicer for the `http-api` in the meanwhile grin

A double negative :eyes: So it _can_ wait? :joy:

FintanH:
Cool! Well if you'd like to help out on the `git-ref-format` migration, always happy to look at PRs or issues :)

keepsimple1:
@sebastinez I would love to get your comments, especially regarding `revision_branches` changes in the attached PR #44 .

sebastinez:
> @sebastinez I would love to get your comments, especially regarding `revision_branches` changes in the attached PR #44 .

Hey @keepsimple1 honestly not really able to give you a review on the whole PR, but the `revision_branches` function looks great, and seems to fix this issue

PR #44, opened 2022-11-02
WIP: radicle-surf: use git-ref-format

keepsimple1:
This is to solve issue #15 and hopefully also address issue #13.  The current diff is a very small proof-of-concept:  `pattern2` macro that produces `const` patterns.  Also added `branches()` method that uses this macro. Please let me know if looks ok and can move forward.


git-ref-format/macro/src/lib.rs
```diff
@@ -166,3 +166,29 @@ pub fn pattern(input: TokenStream) -> TokenStream {
...
+    let parsed: Result<&PatternStr, Error> = val.as_str().try_into();
+    match parsed {
+        Ok(safe) => {
+            let safe: &str = safe.as_str();
+            let expand = quote! {
+                unsafe {
+                    use ::std::mem::transmute;
+                    use ::git_ref_format::refspec::PatternStr;
+
+                    transmute::<_, &PatternStr>(#safe)
```
keepsimple1, line 185: Compared to `pattern`, we return a `&PatternStr` here, so that we can define `const` patterns.

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -75,6 +76,9 @@ impl<'a> From<&'a git2::Repository> for RepositoryRef<'a> {
     }
 }
 
+/// A const ref pattern for local branches.
+pub const REF_LOCAL_BRANCH: &PatternStr = pattern2!("refs/heads/*");
```
keepsimple1, line 80: I felt it's good to define such const patterns. Does it make sense?

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -84,6 +88,13 @@ impl<'a> RepositoryRef<'a> {
             .transpose()
     }
 
+    /// Returns the list of branches that match `pattern`.
+    pub fn branches(&self, pattern: &PatternStr) -> Result<Vec<String>, Error> {
```
keepsimple1, line 92: This method argument is inspired by the discussion in issue #13 .

git-ref-format/macro/src/lib.rs
```diff
@@ -166,3 +166,29 @@ pub fn pattern(input: TokenStream) -> TokenStream {
         },
     }
 }
+
+#[proc_macro_error]
+#[proc_macro]
+pub fn pattern2(input: TokenStream) -> TokenStream {
```
FintanH, line 172: Few things here:
* Can you add documentation, please? -- you can copy and edit the one from `pattern`
* I'm not into this naming :) I'd suggest `pattern_str`. It'd break compatibility but I think we should then call the one above `pattern_string`
* Can you add tests, please?

radicle-surf/src/vcs/git/repo.rs
FintanH: Ya, that's cool :) I'd suggest calling it `REFS_HEADS`

radicle-surf/src/vcs/git/repo.rs
FintanH: This permits any `PatternStr`. I'd expect this function only to return `refs/heads`.

This is why I mentioned a `BranchPattern` in my document. I thought it would look something like this:
```rust
pub struct BranchPattern {
  branch: PatternString
}

impl From<BranchPattern> for PatterString {
  fn from(pat: BranchPattern) -> Self {
    refname!("refs/heads").to_pattern(pat.branch)
  }
}
```

FintanH, line 92: Also let's start moving away from `String` return types and start using `Qualified` -- since we expect references that are `refs/heads`, so they are qualified.

FintanH (commented):
Generally, this is the idea yes :) Left some feedback

radicle-surf/src/vcs/git/repo.rs
keepsimple1: Just curious,  is it because our users prefer `Heads` over `Local branches` in previous experience? or some other reason?

radicle-surf/src/vcs/git/repo.rs
keepsimple1: Agreed. I didn't clarify that I hadn't get around to take care of the return type. 

> I'd expect this function only to return refs/heads

It could also return remote branches though.

git-ref-format/macro/src/lib.rs
keepsimple1: Sure. I was trying to use `pattern` but found that the returned type is not `const` so I added another one just to test things out. 

I was actually thinking to replace `pattern` with this one as it seems that most use cases work well with `const`. I didn't because I wanted to get initial feedback, and also not sure if we have to keep backward compatibility for the `pattern` macro.

What do you think?

radicle-surf/src/vcs/git/repo.rs
keepsimple1: Yes.

radicle-surf/src/vcs/git/repo.rs
FintanH: It's because `REF_LOCAL_BRANCH` doesn't really describe to me what it is, I suppose most people know what local is, but heads might be more prevalent and it's also singular where it should be plural. `REFS_HEADS` literally says what it is. Another suggestion would be `LOCAL_BRANCHES` :)

git-ref-format/macro/src/lib.rs
FintanH: I think we should keep `pattern` because we should allow users to define patterns that are non-`const`. Since it's the early days of this crate we don't need to keep compatibility, we'll just bump the version to `0.2.0` :)

radicle-surf/src/vcs/git/repo.rs
FintanH: > It could also return remote branches though.

Mmm, I see. I would say that maybe we have two separate functions `heads` and `remotes`. They'll both end up returning iterators of `Qualified` so a user can do:
```rust
let branches = surf::heads(surf::REFS_HEADS).chain(surf::remotes(remote_pattern)?);
```

keepsimple1: We probably won't need to pass in `surf::REFS_HEADS` for the function `surf::heads()` as it's supposed to return only the heads anyway.  (Same goes for passing `remote_pattern` to surf::remotes()).

FintanH: I don't understand. Are you saying you wouldn't allow any filtering in the `heads` or `remotes` function?

keepsimple1: I was saying `heads` doesn't need filtering. What are the examples of filters for `heads`?  For `remotes` yes it would need filtering.

FintanH: Mmmmm I see, I was thinking if one wanted to filter by a prefix. Say if I organised my branches under `fintohaps` and just wanted to look at those so I'd pass `fintohaps/*`

keepsimple1: Got it. Will consider that.

git-ref-format/macro/src/lib.rs
keepsimple1: removed this change.

radicle-surf/src/vcs/git/repo.rs
keepsimple1: This change is removed.

git-ref-format/macro/src/lib.rs
keepsimple1: This change is removed.

radicle-surf/src/vcs/git/repo.rs
keepsimple1: I moved to return an iterator of `Branch`. The `Branch` type could be refactored separately.

radicle-surf/Cargo.toml
```diff
@@ -33,6 +33,7 @@ either = "1.5"
 nom = "6"
 nonempty = "0.5"
 radicle-git-ext = { path = "../radicle-git-ext", features = ["serde"] }
+git-ref-format = { path = "../git-ref-format" }
```
FintanH, line 36: Could you also add the versions for `git-ref-format` and `radicle-git-ext`? This will be handy when it comes to publishing `radicle-surf` :)

radicle-surf/src/vcs/git/tag.rs
```diff
@@ -59,6 +53,11 @@ impl TagName {
     pub fn name(&self) -> &str {
         &self.0
     }
+
+    /// Returns the full ref name of the tag.
+    pub fn refname(&self) -> String {
```
FintanH, line 58: Why doesn't this return `RefString`?

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -75,6 +75,152 @@ impl<'a> From<&'a git2::Repository> for RepositoryRef<'a> {
...
+impl<T> Glob<T> {
+    /// Returns the globs.
+    pub fn globs(&self) -> Vec<&str> {
+        self.globs.iter().map(|g| g.as_str()).collect()
+    }
+}
+
+impl<Namespace> Glob<Namespace> {
+    /// Creates a `Glob` for namespaces.
+    pub fn namespaces(glob: &str) -> Result<Self, Error> {
```
FintanH, line 93: I think `&str` is a bad type to accept here. If we use `PatternStr` or even `AsRef<PatternStr>`, then you don't have to return an `Error`. The user of this function can then freely use `pattern!` or do the validation themselves and handle the error on their own turf.

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -75,6 +75,152 @@ impl<'a> From<&'a git2::Repository> for RepositoryRef<'a> {
...
+}
+
+impl<T> Glob<T> {
+    /// Returns the globs.
+    pub fn globs(&self) -> Vec<&str> {
+        self.globs.iter().map(|g| g.as_str()).collect()
+    }
+}
+
+impl<Namespace> Glob<Namespace> {
```
FintanH, line 91: A nice interface for this would be to have a `FromIterator` which allows you to use `collect`.

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -75,6 +75,152 @@ impl<'a> From<&'a git2::Repository> for RepositoryRef<'a> {
...
+    type Item = Result<Tag, Error>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        while self.current < self.references.len() {
+            match self.references.get_mut(self.current) {
+                Some(refs) => match refs.next() {
+                    Some(res) => {
+                        return Some(res.map_err(Error::Git).and_then(|r| {
+                            let b = Tag::try_from(r)?;
+                            Ok(b)
```
FintanH, line 184: nit: could this be just `Tag::try_from(r)`? If not, we could simplify it to `Ok(Tag::try_from(r)?)`

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -84,56 +230,50 @@ impl<'a> RepositoryRef<'a> {
...
+        for glob in pattern.globs.iter() {
+            let new_set = self
+                .repo_ref
+                .references_glob(glob.as_str())?
+                .map(|reference| {
+                    reference
+                        .map_err(Error::Git)
+                        .and_then(|r| Namespace::try_from(r).map_err(|_| Error::EmptyNamespace))
+                })
+                .collect::<Result<HashSet<Namespace>, Error>>()?;
```
FintanH, line 273:
```suggestion
                .collect::<Result<BTreeSet<Namespace>, Error>>()?;
```

I tend to prefer `BTreeSet` because I think the hasher is a bit expensive, iirc

radicle-surf/src/revision.rs
```diff
@@ -79,26 +77,38 @@ pub enum Revision<P> {
...
+                    None => git::Branch::local(name).refname(),
+                };
+                repo.refname_to_oid(&refname)
+            },
+            Revision::Sha { sha } => Ok(*sha),
         }
     }
 }
 
+impl<P> ToCommit for &Revision<P>
```
FintanH, line 103: I'm curious, do you plan on removing the `Revision` type?

FintanH (commented):
This is looking good.

I expected the `name`s for `Branch`, `Tag`, etc. to also be using `git-ref-format` though. Is that a follow-up task?

keepsimple1:
> I expected the `name`s for `Branch`, `Tag`, etc. to also be using `git-ref-format` though. Is that a follow-up task?

Yes, I wanted to take a bit more time and iterate on that.

radicle-surf/src/revision.rs
keepsimple1: Yes. I kept it as is for now as I wanted to understand a bit more about the `peer_id` stuff.

radicle-surf/Cargo.toml
keepsimple1: Will do.

radicle-surf/src/vcs/git/repo.rs
keepsimple1: addressed in the new diff.

radicle-surf/src/vcs/git/repo.rs
keepsimple1: I have been thinking about this. I didn't want to re-export `PatternStr` or `PatternString` in `radicle-surf`'s API. 

Maybe it sounds a bit crazy, but if we would like not to return an `Error` (in fact that was my first version locally), I actually would rather to change `Glob` into using `String`: 
```
pub struct Glob<T> {
    globs: Vec<String>, // i.e. not to use PatternString
    glob_type: PhantomData<T>,
}
```

The validation of the `globs` strings will be done lazily when `heads()` etc methods are called. 

In short, I picked a trade-off that not to re-export types from a dependency (i.e. `git-ref-format`) and still validate the inputs in earlier stage.

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -84,56 +230,50 @@ impl<'a> RepositoryRef<'a> {
...
-            .references(self)?
-            .iter()
-            .try_fold(HashSet::new(), |mut acc, reference| {
-                let namespace = Namespace::try_from(reference?)?;
-                acc.insert(namespace);
-                Ok(acc)
-            });
-        Ok(namespaces?.into_iter().collect())
+    /// Returns the list of namespaces that match `pattern`.
+    pub fn namespaces(&self, pattern: &Glob<Namespace>) -> Result<Vec<Namespace>, Error> {
```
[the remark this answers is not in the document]
keepsimple1: In a subtle way, I think `Vec` is more familiar for most users than `BTreeSet`. That was the only reason.

While we're here, what do you think if we return an iterator for namespaces too? That way, internally we can use anything we like, and it is more consistent with `branches()` and `tags()`. I didn't return an iterator because we need to do a `collect` anyway due to the nature of namespaces. But we can argue that `collect` is an implementation detail and should not dictate the return type.

radicle-surf/src/vcs/git/repo.rs
keepsimple1: Will update.

radicle-surf/src/vcs/git/repo.rs
keepsimple1: Sorry I don't understand. Did you mean adding `impl FromIterator` for `Glob`?

radicle-surf/src/vcs/git/tag.rs
keepsimple1: If we return `RefString`, currently we will use `TryFrom` internally and need to handle errors. It will be easier if we change `TagName` itself to be `RefString` later. This is related to your comments below regarding changing `Branch.name` type.

FintanH:
> > I expected the `name`s for `Branch`, `Tag`, etc. to also be using `git-ref-format` though. Is that a follow-up task?
> 
> Yes, I wanted to take a bit more time and iterate on that.

Ok, good to know :+1:

radicle-surf/src/revision.rs
FintanH: The `peer_id` is a remnant of the old protocol. A reference for a remote peer would be of the form:
```
refs/namespaces/<project>/refs/remotes/<peer_id>/...
```

This ended up being a bit haphazard in the long run and it'll be unlikely that we'll see this in the new protocol. A project is scoped to a single git repository and the peer is scoped to a namespace:
```
refs/namespaces/<peer>/
  refs/heads/...
  refs/tags/...
```
There won't be any remotes because those git repositories won't have a `remote` entry in their config files.

But it's still worth supporting `refs/remotes` (without a namespace) for local working copies.

radicle-surf/src/vcs/git/repo.rs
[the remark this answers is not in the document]
FintanH: > In a subtle way, I think `Vec` is more familiar for most users than `BTreeSet`. That was the only reason.

What makes you say that? `BTreeSet` is a well-known structure in Rust and it's in the standard library. It seems a waste to collect into it and then turn that into a `Vec`, especially if the semantics are meant to say that the namespaces returned are unique.

> While we're here, what do you think if we return an iterator for namespaces too? That way, internally we can use anything we like, and it is more consistent with `branches()` and `tags()`. I didn't return an iterator because we need to do a `collect` anyway due to the nature of namespaces. But we can also argue that `collect` is an implementation detail and should not dictate the return type.

Yup, that sounds good! Iterator all the things :)

radicle-surf/src/vcs/git/repo.rs
FintanH: Sorry, a bit of a red herring. I don't think we should implement `FromIterator` for `Glob` because that would break encapsulation. However, it would make sense if we wrote this code more like this:
```rust
use git_ref_format::refname;

pub struct NamespacePattern {
    pub pattern: PatternString,
}

impl From<PatternString> for NamespacePattern {
    fn from(pattern: PatternString) -> Self {
        Self { pattern }
    }
}

impl NamespacePattern {
    pub fn refspec(self) -> PatternString {
        refname!("refs/namespaces").to_pattern(self.pattern)
    }
}

pub struct Glob<T> {
    globs: Vec<T>,
}

impl<P: Into<NamespacePattern> FromIterator<P> for Globber<NamespacePattern> {
    fn from_iter<T: IntoIterator<Item = P>>(iter: T) -> Self {
        Globber {
            globs: iter.into_iter().map(NamespacePattern::from).collect()
        }
    }
}
```
And then you could have the same for `RemotePattern`, `HeadsPattern`, and `TagPattern`. `BranchPattern` could be defined as an `enum` in terms of both `RemotePattern` and `HeadsPattern`.

To build a `Glob` then you could do:
```rust
let namespaces: Glob<Namespace> = vec![
    pattern!("refs/heads/han*"), 
    pattern!("refs/heads/finto/*")].into_iter().collect();
```

radicle-surf/src/vcs/git/repo.rs
FintanH: I think it's fine to expose `git-ref-format` as part of our API. I stated why it's a beneficial library in the issue thread. I wouldn't say we need to re-export it though.

radicle-surf/src/vcs/git/tag.rs
FintanH: Ya, you're right. The internal names need to be changed first :+1:

radicle-surf/src/vcs/git/repo.rs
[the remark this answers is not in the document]
keepsimple1: > What makes you say that? BTreeSet is a well-known structure in Rust and it's in the standard library. 

One thing is that: you don't need to import `Vec` to use it in Rust, but you need to import `BTreeSet` (`use std::collections::BTreeSet`) to use it.  It's a minor thing, but it tells something. 

I will work on returning an iterator.

radicle-surf/src/vcs/git/repo.rs
keepsimple1: I see. Following offline discussions, currently I would like to keep `git-ref-format` types as an implementation detail, hence, not to expose them via API parameters.  If we decide to re-export or require surf users to use `git-ref-format` types, then we can change and support `From<PatternString>` like above, or directly to `Glob<T>`. I will leave that to another PR.

radicle-surf/src/vcs/git/repo.rs
keepsimple1: > I think it's fine to expose git-ref-format as part of our API. I stated why it's a beneficial library in the issue thread. I wouldn't say we need to re-export it though.

It is certainly one way to go about it. IMO, I would like to keep `git-ref-format` types as implementation details for as long as we could. My experience is that exposing types from a dependency has more downsides and upsides in most cases, so I'm a bit cautious.

PR #45, opened 2022-11-04
radicle-git-ext: remove multihash and minicbor

### LINKED ISSUE #42: Multihash dependency isn't needed
We have a `multihash` dependency that is used to convert an `Oid` into a multihash. I think we could remove this and drop the dependency.

FintanH:
Multihash was deemed as not needed for compatibility since it is possible to detect a change from SHA1 to SHA256 based on their encoded lengths. The YAGNI principle is executed and multihash (and minicbor while we're here) are removed.

This also had an effect on `Urn` which relied heavily on encoding the value using multihash. The `urn` and `namespace` modules were removed from radicle-git-types, since they were likely not going to be used anyway.

Fixes #42


cloudhead (dismissed):
Yess

Issue #9, opened 2022-09-19
surf: `Repository::new` or similar should search up the parent directories.


keepsimple1:
An enhancement to consider:  `Repository::new` or a different method to use `git2::Repository::open_ext` so that it can automatically search up the parent directories.

_Originally posted by @keepsimple1 in https://github.com/radicle-dev/radicle-git/pull/3#discussion_r966439815_

PR #50, opened 2022-11-11
search parent directories in Repository::open

keepsimple1:
Fix issue #9 .

- Replaces the existing Repository::new() with ::open() that supports searching parent directories. If we deem that `::new()` should be kept, we can add it back that simply calls `open()` internally.

### DIFF
--- a/radicle-surf/src/vcs/git/repo.rs
+++ b/radicle-surf/src/vcs/git/repo.rs
@@ -536,21 +536,28 @@ impl<'a> std::fmt::Debug for RepositoryRef<'a> {
 }
 
 impl Repository {
-    /// Open a git repository given its URI.
+    /// Open a git repository given its exact URI.
     ///
     /// # Errors
     ///
     /// * [`Error::Git`]
-    pub fn new(repo_uri: impl AsRef<std::path::Path>) -> Result<Self, Error> {
+    pub fn open(repo_uri: impl AsRef<std::path::Path>) -> Result<Self, Error> {
         git2::Repository::open(repo_uri)
+            .map(Repository)
+            .map_err(Error::from)
+    }
+
+    /// Attempt to open a git repository at or above `repo_uri` in the file
+    /// system.
+    pub fn discover(repo_uri: impl AsRef<std::path::Path>) -> Result<Self, Error> {
+        git2::Repository::discover(repo_uri)
             .map(Repository)
             .map_err(Error::from)
     }
 
     /// Since our operations are read-only when it comes to surfing a repository
     /// we have a separate struct called [`RepositoryRef`]. This turns an owned
-    /// [`Repository`], the one returned by [`Repository::new`], into a
-    /// [`RepositoryRef`].
+    /// [`Repository`] into a [`RepositoryRef`].
     pub fn as_ref(&'_ self) -> RepositoryRef<'_> {
         RepositoryRef { repo_ref: &self.0 }
     }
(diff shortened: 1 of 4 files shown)


radicle-surf/src/vcs/git/repo.rs
```diff
@@ -536,21 +537,40 @@ impl<'a> std::fmt::Debug for RepositoryRef<'a> {
...
+    /// Returns an error if cannot find a valid git repo to open.
+    pub fn open<P>(repo_uri: P, up_until: Option<P>) -> Result<Self, Error>
+    where
+        P: AsRef<std::path::Path>,
+    {
+        match up_until {
+            None => git2::Repository::open(repo_uri)
+                .map(Repository)
+                .map_err(Error::from),
+            Some(p) => git2::Repository::open_ext(
```
FintanH, line 561: Any reason you opted to use `open_ext` over [`discover`](https://docs.rs/git2/0.15.0/git2/struct.Repository.html#method.discover)?

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -536,21 +537,40 @@ impl<'a> std::fmt::Debug for RepositoryRef<'a> {
...
-    /// # Errors
+    /// `up_until`: if it is `None`, `repo_uri` must be a valid git repo.
+    /// Otherwise, we will search parent directories until we find a
+    /// valid git repo, or reach `up_until`. Note that `up_until` itself will
+    /// not be searched.
     ///
-    /// * [`Error::Git`]
-    pub fn new(repo_uri: impl AsRef<std::path::Path>) -> Result<Self, Error> {
-        git2::Repository::open(repo_uri)
+    /// A quirk: if `up_until` matches the starting directory (`repo_uri`),
```
FintanH, line 547: To prevent this quirk, we could check if `repo_uri == p` in the match guard, and use `open` if it is the case.

radicle-surf/src/vcs/git/repo.rs
```diff
@@ -536,21 +537,40 @@ impl<'a> std::fmt::Debug for RepositoryRef<'a> {
...
+    /// See also: <https://github.com/rust-lang/git2-rs/issues/892>
+    ///
+    /// Returns an error if cannot find a valid git repo to open.
+    pub fn open<P>(repo_uri: P, up_until: Option<P>) -> Result<Self, Error>
+    where
+        P: AsRef<std::path::Path>,
+    {
+        match up_until {
+            None => git2::Repository::open(repo_uri)
+                .map(Repository)
```
FintanH, line 559: You could do the shared `map` and `map_err` outside the match to save writing it twice

radicle-surf/t/src/git.rs
```diff
@@ -17,7 +17,12 @@ const GIT_PLATINUM: &str = "../data/git-platinum";
 #[test]
 // An issue with submodules, see: https://github.com/radicle-dev/radicle-surf/issues/54
 fn test_submodule_failure() {
-    let repo = Repository::new("../..").unwrap();
+    // Search up until the home directory for the repo.
+    let repo = Repository::open(".", Some("~")).unwrap();
```
FintanH, line 21: Are you positive the `~`  always means home directory?

radicle-surf/src/vcs/git/repo.rs
keepsimple1: I chose `open_ext` as it allows the user to set a ceiling_dir.  `discover` does not.

I don't know which way works better in real world use cases. I am okay with your suggestion.

radicle-surf/t/src/git.rs
keepsimple1: no longer needed in the updated diff.

FintanH (approved):
LGTM :)

PR #56, opened 2022-11-16
radicle-surf: iterator for names

FintanH:
The `tag_names` and `branch_names` functions use the `Tags` and
`Branches` functions which require to go through their `TryFrom`
instances. This can be expensive, in the case of `Tags` especially.

Add `TagNames` and `BranchNames` iterators that only return the names
of the tags and branches, respectively.

This requires to change the definition of Revisions, where the
NonEmpty branch list is now foregone and there references are
collected in a BTreeSet to ensure ordering.

### DIFF
--- a/radicle-surf/src/git/repo.rs
+++ b/radicle-surf/src/git/repo.rs
@@ -23,7 +23,7 @@ use std::{
 };
 
 use directory::{Directory, FileContent};
-use git_ref_format::{refspec::QualifiedPattern, Qualified, RefString};
+use git_ref_format::{refspec::QualifiedPattern, Qualified};
 use radicle_git_ext::Oid;
 use thiserror::Error;
 
@@ -50,6 +50,8 @@ use crate::{
 
 pub mod iter;
 pub use iter::{Branches, Namespaces, Tags};
+
+use self::iter::{BranchNames, TagNames};
 
 /// Enumeration of errors that can occur in operations from [`crate::git`].
 #[derive(Debug, Error)]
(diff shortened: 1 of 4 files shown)


radicle-surf/src/git/branch.rs
```diff
@@ -219,6 +237,18 @@ impl Remote {
...
+    /// Return the short `Remote` refname,
+    /// e.g. `fix/ref-format`.
+    pub fn short_name(&self) -> &RefString {
+        &self.name
+    }
+
+    /// Return the remote of the `Remote`'s refname,
+    /// e.g. `origin`.
+    pub fn remote(&self) -> Component {
+        Component::from_refstring(self.remote.clone()).expect("BUG: remote should be a Component")
```
keepsimple1, line 249: We should not panic here.  It's trade-off and I would suggest return `&RefString`  for this method.

keepsimple1 (commented):
Looks good to me, thanks! only minor comment inline.

keepsimple1:
nit: could you please also add a brief description for this PR, or a link to an issue (though I don't think there is an issue for this PR)?

FintanH:
> nit: could you please also add a brief description for this PR, or a link to an issue (though I don't think there is an issue for this PR)?

I described what happened in the commits, but can copy that

radicle-surf/src/git/branch.rs
FintanH: If the `remote` is not a `Component` then `Remote` was constructed incorrectly. The reason I wanted to return a `Component` is that it mirrors the constructor.

But happy to change it to `RefString`.

PR #61, opened 2022-11-18
radicle-surf: make directory and file contents generic

FintanH:
Change the inner value of a File and Directory to be generic. This allows the inner value to be an Oid, if it's not resolved. If the content are resolved they will be FileContent & Entries respectively.

This allows the writing of nice helper functions such as `fold`, `try_fold`, and `traverse`.

### DIFF
--- a/radicle-surf/src/file_system/directory.rs
+++ b/radicle-surf/src/file_system/directory.rs
@@ -20,55 +20,114 @@
 //! A `Directory` is expected to be a non-empty tree of directories and files.
 //! See [`Directory`] for more information.
 
-use crate::{
-    file_system::{error::LabelError, path::*, Error},
-    git::{self, Repository, Revision},
-};
-use git2::Blob;
-use radicle_git_ext::Oid;
 use std::{
     collections::BTreeMap,
-    convert::{Infallible, TryFrom},
+    convert::{Infallible, Into},
     path,
 };
 
-/// Represents a `file` in a git repo.
+use git2::Blob;
+use radicle_git_ext::{is_not_found_err, Oid};
+use radicle_std_ext::result::ResultExt as _;
+
+use crate::{
+    file_system::{path::*, Error},
+    git::{Repository, Revision},
+};
+
+pub mod error {
+    use thiserror::Error;
+
+    #[derive(Debug, Error, PartialEq)]
+    pub enum Directory {
+        #[error(transparent)]
+        Git(#[from] git2::Error),
+        #[error(transparent)]
+        Entry(#[from] Entry),
+        #[error(transparent)]
+        File(#[from] File),
+    }
+
+    #[derive(Debug, Error, PartialEq, Eq)]
+    pub enum Entry {
+        #[error("the entry name was not valid UTF-8")]
+        Utf8Error,
+        #[error(transparent)]
+        Label(#[from] super::Error),
+    }
+
+    #[derive(Debug, Error, PartialEq)]
+    pub enum File {
+        #[error(transparent)]
+        Git(#[from] git2::Error),
+    }
+}
+
+/// A `File` in a git repository.
+///
+/// The representation is lightweight and contains the [`Oid`] that
+/// points to the git blob which is this file.
+///
+/// The name of a file can be retrieved via [`File::name`].
+///
+/// The [`FileContent`] of a file can be retrieved via
+/// [`File::content`].
 #[derive(Clone, PartialEq, Eq, Debug)]
 pub struct File {
     pub(crate) name: Label,
-    pub(crate) oid: Oid,
+    pub(crate) id: Oid,
 }
 
 impl File {
-    /// Create a new `File`.
-    pub fn new(name: String, oid: Oid) -> Self {
-        File {
+    /// The name of this `File`.
+    pub fn name(&self) -> &str {
+        self.name.as_str()
+    }
+
+    /// The object identifier of this `File`.
+    pub fn id(&self) -> Oid {
+        self.id
+    }
+
+    /// Create a new `File` with the `name` and `id` provided.
+    ///
+    /// The `id` must point to a `git` blob.
+    pub fn new(name: String, id: Oid) -> Self {
+        Self {
             name: Label {
                 label: name,
                 hidden: false,
             },
-            oid,
+            id,
         }
     }
 
-    /// Returns the file name reference.
-    pub fn name(&self) -> &str {
-        self.name.as_str()
+    /// Get the [`FileContent`] for this `File`.
+    ///
+    /// # Errors
+    ///
+    /// This function will fail if it could not find the `git` blob
+    /// for the `Oid` of this `File`.
+    pub fn content<'a>(&self, repo: &'a Repository) -> Result<FileContent<'a>, error::File> {
+        let blob = repo.git2_repo().find_blob(self.id.into())?;
+        Ok(FileContent { blob })
     }
 }
 
-/// Represents the actual content of a file.
+/// The contents of a [`File`].
+///
+/// To construct a `FileContent` use [`File::content`].
 pub struct FileContent<'a> {
     blob: Blob<'a>,
 }
 
 impl<'a> FileContent<'a> {
-    /// Returns the file content as a byte slice.
+    /// Return the file contents as a byte slice.
     pub fn as_bytes(&self) -> &[u8] {
         self.blob.content()
     }
 
-    /// Returns the size of file
+    /// Return the size of the file contents.
     pub fn size(&self) -> usize {
         self.blob.size()
     }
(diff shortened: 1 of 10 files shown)


radicle-surf/src/file_system/directory.rs
```diff
@@ -80,98 +138,181 @@ impl<'a> FileContent<'a> {
...
-/// Represents an entry in a directory.
+/// `Entriess` is an enumeration of what a [`Directory`] can contain
+/// and is used for when we are iterating through a [`Directory`].
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub enum DirectoryEntry {
-    /// When the entry is a file.
-    File(File),
-    /// When the entry is a directory.
-    Directory(Directory),
+pub enum Entry<F, D> {
```
keepsimple1, line 155: IMO, `F` and `D` are not coherent with each other. They can the same or different types and that doesn't tell the users much. I understand this is the consequence of defining both `File` and `Directory` as generic, but it just feels too much cognitive load.

radicle-surf/src/file_system/directory.rs
```diff
@@ -80,98 +138,181 @@ impl<'a> FileContent<'a> {
...
+}
+
 /// A `Directory` is a set of entries of sub-directories and files, ordered
 /// by their unique names in the alphabetical order.
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct Directory {
+pub struct Directory<T> {
     pub(crate) name: Label,
-    pub(crate) oid: Oid,
+    pub(crate) entries: T,
```
keepsimple1, line 193: It seems to me that the naming of `entries` is to accommodate the directory contents in `T`,  but then it feels a bit odd when `T` is a single `Oid`.

radicle-surf/src/file_system/directory.rs
```diff
@@ -80,98 +138,181 @@ impl<'a> FileContent<'a> {
...
-impl Directory {
-    /// Creates a directory given `name` and `oid`, with empty contents.
-    pub fn new(name: Label, oid: Oid) -> Self {
-        Directory { name, oid }
+impl<F, D> Directory<Entries<F, D>> {
+    pub fn entries(&self) -> impl Iterator<Item = &Entry<F, D>> {
+        self.entries.iter()
+    }
+
+    pub fn fold<B, G>(&self, initial: B, g: G) -> B
```
keepsimple1, line 201: Having `B`, `G` on top of `F`, `D` is killing me :( . How do you add doc comments for this method?

keepsimple1 (commented):
Overall this diff feels overengineered to me. It seems that `Oid` and `FileContent`, `Entries<F, D>` are squeezed into the same type, not naturally fitting in.

What is the actual problem we are trying to solve?

FintanH:
> Overall this diff feels overengineered to me. It seems that `Oid` and `FileContent`, `Entries<F, D>` are squeezed into the same type, not naturally fitting in.
> 
> What is the actual problem we are trying to solve?

I would say that this is a common pattern in functional programming languages or at least languages that give you generics. The idea is that we have two representations of a `File` and a `Directory`. The first for both is the one that uses the `Oid` as a pointer to the useful information, `FileContent` and `Entries`, respectively. The second is the resolution of both of these types. This could be encoded with 4 types, like:
```rust
pub struct PointerFile { name: String, oid: Oid }
pub struct PointerDirectory { name: String, oid: Oid }

pub struct File { name: String, content: FileContent }
// Note that Entries either has to be `File` or `Directory` OR we also introduce `PointerEntries`
pub struct Directory { name: String, entries: Entries }

impl PointerFile {
  pub fn resolve(&self, repo: &Repository) -> File {
      /* code */
  }
}

/* etc. */
```

Now instead of duplicating those types, we abstract over the differing parts `Oid` and `FileContent`/`Entries` to get a single representation that can be transformed via the functions I called `resolve`. We can then organise functions into their completely generic form, i.e. `name`, and their specifics, i.e. `resolve`, `size`, `entries`, `fold`, etc. 

So I would argue that they are squeezed into the same type because they are naturally fitting, they allow the conversion from the one to other -- `Oid -> Entries`, `Oid -> FileContent`.

radicle-surf/src/file_system/directory.rs
FintanH: Well if `T` is a single `Oid`, the `Oid` _represents_ the entries because it's the pointer to them :)

radicle-surf/src/file_system/directory.rs
FintanH: I've added (hopefully) more clear documentation on all the generic aspects. Let me know if that helps :)

FintanH:
I'm exploring the `src/tree.rs` code at the moment I'm noticing another benefit to having this be generic. A [`TreeEntry`] is basically our `directory::Entry`, where `Info` is capturing the `enum` of it, but it also has this `last_commit` field. We could trivially remove `TreeEntry` and replace it with `directory::Entry` with something like the following:
```rust
pub struct Info { last_commit: commit::Header, content: FileContent } 
pub type TreeEntry = directory::Entry<Info, Directory<TreeEntry>>;
```

keepsimple1:
> > Overall this diff feels overengineered to me. It seems that `Oid` and `FileContent`, `Entries<F, D>` are squeezed into the same type, not naturally fitting in.
> > What is the actual problem we are trying to solve?
> 
> I would say that this is a common pattern in functional programming languages or at least languages that give you generics. 

Is there a name for this pattern?

Functional programming aside, I think there are a couple of practical problems in this case:

1. When `File<T>` and `Directory<E>` are `File<FileContent>` and `Directory<Entries<F, D>>` , we will be holding potential large objects in memory with potential large hierarchy. This leads back to the original scale issue I talked about when introducing `Oid` in the first place. 

And now, these potential large objects are more difficult to spot because they are "hidden" behind the generics.

2.  When `File` and `Directory` are resolved, they no longer have `Oid`,  which is awkward, and IMO a problem.  This means `File` sometimes has its `Oid`, and sometimes not. IMO, it's cleaner to think about `File` that always has its `Oid`. The same goes for `Directory`.

keepsimple1:
> ```rust
> pub struct PointerFile { name: String, oid: Oid }
> pub struct PointerDirectory { name: String, oid: Oid }
> 
> pub struct File { name: String, content: FileContent }
> // Note that Entries either has to be `File` or `Directory` OR we also introduce `PointerEntries`
> pub struct Directory { name: String, entries: Entries }
> 
> impl PointerFile {
>   pub fn resolve(&self, repo: &Repository) -> File {
>       /* code */
>   }
> }
> 
> /* etc. */
> ```
> 
> Now instead of duplicating those types, we abstract over the differing parts `Oid` and `FileContent`/`Entries` to get a single representation that can be transformed via the functions I called `resolve`. 

To clarify, the current code does not have this form of duplication.  `File` has `Oid`,  not `FileContent` in its fields.

keepsimple1:
> I'm exploring the `src/tree.rs` code at the moment I'm noticing another benefit to having this be generic. A [`TreeEntry`] is basically our `directory::Entry`, where `Info` is capturing the `enum` of it, but it also has this `last_commit` field. We could trivially remove `TreeEntry` and replace it with `directory::Entry` with something like the following:

I think you meant `src/object/tree.rs`. I think we might be able to remove it altogether. At this stage of refactoring,

> ```rust
> pub struct Info { last_commit: commit::Header, content: FileContent } 
> pub type TreeEntry = directory::Entry<Info, Directory<TreeEntry>>;
> ```

Is this a recursive type definition? I'm not sure if Rust handles it well. And, I have to say I'm overwhelmed by this complexity.

FintanH:
> Is there a name for this pattern?

I don't think I know of one, and I asked some peers but they didn't have a name for it either. It's more the idea of keeping the common structure but varying the contents. It gives you nice functions like `map` and `traverse` in languages like Haskell :)



> 1. When `File<T>` and `Directory<E>` are `File<FileContent>` and `Directory<Entries<F, D>>` , we will be holding potential large objects in memory with potential large hierarchy. This leads back to the original scale issue I talked about when introducing `Oid` in the first place.

I don't think that's necessarily true. An `Entries<F, D>` is still only one layer of that `Directory`. Technically, you can encode how many layers you have in the types:
```rust
type OneLayer = Directory<Entries<F, D>>;
type TwoLayers = Directory<Entries<F, Entries<F, D>>>;
type AllLayers = Directory<Entries<F, AllLayers>>;
```

The common case is that we tend to work with the equivalent of `OneLayer`.

> 2\. When `File` and `Directory` are resolved, they no longer have `Oid`,  which is awkward, and IMO a problem.  This means `File` sometimes has its `Oid`, and sometimes not. IMO, it's cleaner to think about `File` that always has its `Oid`. The same goes for `Directory`.

That's fair but also a solvable problem. We could create a struct for holding the `Oid` and the `FileContent`. I think this improves things because in the previous iteration your `File` with the `Oid` is separated from the `FileContent` so you have to hold two variables for them which feels awkward, personally, but not a deal breaker :)

> To clarify, the current code does not have this form of duplication. `File` has `Oid`, not `FileContent` in its fields.

Sure, that was for demonstration mostly. It was showing how we could have all those types but using a single struct :)





> I think you meant `src/object/tree.rs`. I think we might be able to remove it altogether.
> 
> > ```rust
> > pub struct Info { last_commit: commit::Header, content: FileContent } 
> > pub type TreeEntry = directory::Entry<Info, Directory<TreeEntry>>;
> > ```
> 
> Is this a recursive type definition? I'm not sure if Rust handles it well. And, I have to say I'm overwhelmed by this complexity.

Aye, sorry it is that file! And yes, I'm going to look at removing that, or at least reusing `Directory` to define it.

And yes it is a recursive definition, I think I have to correct myself since it would be `directory::Entry<Info, TreeEntry>`.


All that being said, since you're working on `radicle-surf` as well, I'd rather not overwhelm you with complexity if you think this is too much. I can go back to the drawing board.

keepsimple1 (approved):
Thanks for the updates. LGTM!

PR #62, opened 2022-11-18
radicle-surf: remove file_system::path

FintanH:
The file_system types, Label & Path, are overengineered. It's more natural to use the Path and PathBuf types provided by the std library.

This replaces any mention of Label with String and any use of the old Path with either std::path::Path or std::path::PathBuf, depending on what kind of ownwership is needed.

Anywhere a "root" directory is needed `PathBuf::new()` is used, i.e. an empty `PathBuf`. Otherwise `Path::new` can used to pass directory, or in fact a &str can also be used since it implements AsRef<Path>.

The `File` and `Directory` types also contain there absolute paths, which can be retrieved while traversing the Tree.

Note the special file name `"special/faux\\path"` requires the use of an escaping string construction, i.e. `r"special/faux\\path"`.


radicle-surf/src/file_system/directory.rs
```diff
@@ -66,7 +61,8 @@ pub mod error {
 /// Represents a `file` in a git repo.
 #[derive(Clone, PartialEq, Eq, Debug)]
 pub struct File<T> {
-    pub(crate) name: Label,
+    pub(crate) name: String,
+    pub(crate) path: PathBuf,
```
keepsimple1, line 65: 1. What is the reason to need a `path`?
2. Does `path` include name?  If yes, is `name` a duplicated info?2
3. Is `path` absolute in the sense of OS file system? or only relative to the git root? (I understand normally a path is absolute in OS file system, but wanted to confirm).

keepsimple1 (commented):
Overall looks good to me, thanks! Few comments inline.

radicle-surf/src/file_system/directory.rs
keepsimple1: no problem, I will review again once the documentation is updated. For adding the path, I'm ok to have it but still don't know if it's required. And between `name` and `path`, could you please document which one is the source of truth and which is derived data?

radicle-surf/src/file_system/directory.rs
```diff
@@ -74,11 +69,32 @@ pub mod error {
 /// [`File::content`].
 #[derive(Clone, PartialEq, Eq, Debug)]
 pub struct File {
-    pub(crate) name: Label,
-    pub(crate) id: Oid,
+    /// The name of the file.
+    name: String,
+    /// The relative path of the file, not including the `name`,
+    /// in respect to the root of the git repository.
+    prefix: PathBuf,
```
keepsimple1, line 76: 👍 Thanks!

keepsimple1 (dismissed):
It's looking good! Thanks.

PR #63, opened 2022-11-18
radicle-surf: organise tests

FintanH:
Reorganise the tests into module files, rather than all being in a single file. No tests were changed in this patch.

The `serialize` flag was also not enabled in `test/Cargo.toml`. This is enabled to ensure those tests are run moving forward.


keepsimple1 (dismissed):
LGTM!

PR #65, opened 2022-11-18
git-ref-format: remove link-literals feature

### LINKED ISSUE #58: Remove `link` feature?
I noticed we still had some functionality feature-flagged under `link` in git-ref-format (and maybe other places?); I'd suggest this can be removed now as it's not in use as far as I know.

FintanH:
Since this package is not intended for radicle-link anymore, the feature flag and the literals are removed.

FIxes #58


keepsimple1 (approved):
LGTM!

Issue #67, opened 2022-11-21
radicle-surf: RepositoryRef is not necessary


keepsimple1:
The reference type [`RepositoryRef`](https://github.com/radicle-dev/radicle-git/blob/d3115a22158c8395705babefdc89049f7510d32d/radicle-surf/src/git/repo.rs#L113) adds more boilerplate code than it's worth. We should be able to remove it and use `&Repository` directly, and make the API surface more concise.

PR #68, opened 2022-11-21
Remove RepositoryRef type

keepsimple1:
Fix issue #67 .  It is like merging `String` and `&str` (if that was possible ;) ). But this one is possible.

### DIFF
--- a/radicle-surf/README.md
+++ b/radicle-surf/README.md
@@ -1,13 +1,16 @@
 # radicle-surf
 
-A code surfing library for VCS file systems 🏄‍♀️🏄‍♂️
+A code surfing library for Git repository 🏄‍♀️🏄‍♂️
 
 Welcome to `radicle-surf`!
 
-`radicle-surf` is a system to describe a file-system in a VCS world.
-We have the concept of files and directories, but these objects can change over time while people iterate on them.
-Thus, it is a file-system within history and we, the user, are viewing the file-system at a particular snapshot.
-Alongside this, we will wish to take two snapshots and view their differences.
+`radicle-surf` is a library to describe a Git repository as a file system. It
+aims to provide an easy-to-use API to browse a repository via the concept of
+files and directories for any given revision. It also allows the user to diff
+any two different revisions.
+
+One of the use cases would be to create a web GUI for interacting with a Git
+repository (thinking GitHub, GitLab or similar systems).
 
 ## Contributing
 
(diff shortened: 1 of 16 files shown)


radicle-surf/README.md
```diff
@@ -1,13 +1,16 @@
 # radicle-surf
 
-A code surfing library for VCS file systems 🏄‍♀️🏄‍♂️
+A code surfing library for Git repository 🏄‍♀️🏄‍♂️
```
FintanH, line 3:
```suggestion
A code surfing library for Git repositories 🏄‍♀️🏄‍♂️
```

radicle-surf/README.md
```diff
@@ -18,57 +21,8 @@ our [LICENSE](../LICENSE) file.
 
 Join our community disccussions at [radicle.community](https://radicle.community)!
 
+
 # Example
 
 To a taste for the capabilities of `radicle-surf` we provide an example below, but we also
 keep our documentation and doc-tests up to date.
-
-```rust
```
FintanH, line 26: Nice :)

radicle-surf/src/git/repo.rs
```diff
@@ -484,57 +462,39 @@ impl<'a> RepositoryRef<'a> {
...
-    pub fn as_ref(&'_ self) -> RepositoryRef<'_> {
-        RepositoryRef { repo_ref: &self.0 }
-    }
-}
-
-impl<'a> From<&'a Repository> for RepositoryRef<'a> {
-    fn from(repo: &'a Repository) -> Self {
-        repo.as_ref()
+    /// Get a reference to the underlying git2 repo.
+    pub(crate) fn git2_repo(&self) -> &git2::Repository {
```
FintanH, line 490: I have a thought about this and maybe you can tell me what you think.

Having this public to the crate means that there's a kind of abstraction leak of the repo when used in other modules. So I'm thinking that we could have two `impl Repository` definitions. The first is the public API, it only contains `pub fn`. The second is the crate API, so it only contains `pub(crate) fn` -- or just `fn` for the cases of only needing it in this file.

Then for any uses outside of this module, we add a `pub(crate) fn`, for example, `find_commit` is one that I spotted. That way we limit the access to the internal repository handle, no matter what it is.

radicle-surf/src/git/repo.rs
```diff
@@ -99,36 +99,14 @@ pub enum Error {
...
-///
-/// # Construction
-///
-/// Use the `From<&'a git2::Repository>` implementation to construct a
-/// `RepositoryRef`.
-#[derive(Clone, Copy)]
-pub struct RepositoryRef<'a> {
-    pub(crate) repo_ref: &'a git2::Repository,
+pub struct Repository {
+    repo: git2::Repository,
```
FintanH, line 103: nit: I'd rather call this `inner` or `raw`

FintanH (commented):
Generally for this change! Thanks for kicking it off :)

Left some thoughts below.

radicle-surf/src/git/repo.rs
keepsimple1: Sounds good!

keepsimple1: Do you want me to make this change in this PR or in a separate PR?

keepsimple1: Like we discussed, let's merge this PR first and then do the methods separation in another PR. Thanks.

FintanH:
Squashed the 3 commits and merged https://github.com/radicle-dev/radicle-git/commit/c44348fc3c832c64ec84efc2f75d055ac52a576c

keepsimple1:
Thanks for squashing the commit. Sorry that I forgot to do that earlier.

PR #69, opened 2022-11-22
meta: clean up cargo dependencies

FintanH:
The aim of this patch is to solve some of the `cargo deny` warnings. This required updating some dependencies. Some dependencies of radicle-surf were not used, so they were outright removed.

The last_commit benchmark in radicle-surf is irrelevant as we use more efficient methods for getting repository histories, so it has been removed.


keepsimple1 (dismissed):
Nice cleanup! Thanks!

PR #70, opened 2022-11-24
radicle-surf: refactor source logic

FintanH:
`radicle-source` was folded into `radicle-surf` from its own crate, the original decision was to place these modules at the top level. This patch series creates a separated hierarchy, `source`, instead -- see the first commit.

The main goal of this series was to change up the `Blob`, `Tree`, and `Revisions` types. The `Blob` is changed to be a wrapper around a `File`, including the additional fields of `BlobContent` and `Option<commit::Header>`. The `Tree` is similarly changed to be a wrapper around `Directory` -- and `directory::Entry` for `TreeEntry` -- also including the `Option<commit::Header>` field. **N.B.** this changes the serialization of these types and so this is **not a compatible change**.

The `Revisions` type is completely overhauled and renamed to `View`. It was specific to the `radicle-link` protocol and now reflects the Heartwood protocol, as documented. @xphoniex or @sebastinez, do you know if the `Revisions` type was used in the `http-api` at all in the first place?

Finally, the `serialization` feature flag was simplified to `serde`. And I also noticed that the `syntax` flag wasn't carried over, so none of that code was being compiled. That's fixed now.


sebastinez:
Hey @FintanH, thanks for asking for my review, not yet qualified for so many LoC of Rust magic 😅 

wrt your questions:
> do you know if the Revisions type was used in the http-api at all in the first place?

AFAIK we used it at some point, but with the latest moves to `radicle_surf` it seems that we removed any usage of it. 

> And I also noticed that the syntax flag wasn't carried over, so none of that code was being compiled.

we are moving the syntax highlighting over to `radicle-interface`. https://github.com/radicle-dev/radicle-interface/issues/497
so we won't use that in the http-api, maybe we could remove everything under this feature flag also? wdyt? @cloudhead

FintanH:
> Hey @FintanH, thanks for asking for my review, not yet qualified for so many LoC of Rust magic sweat_smile

Thanks @sebastinez :blush: No worries on the review of the code, just wanted to put the non-compatible changes on your radar :)

> > do you know if the Revisions type was used in the http-api at all in the first place?
> 
> AFAIK we used it at some point, but with the latest moves to `radicle_surf` it seems that we removed any usage of it.

And you're not aware of any plan to use it? I'm not sure how the drop-downs operate for delegates in the `http-api`. It would be easy to copy this over to there since it is more specific to `heartwood`.

> > And I also noticed that the syntax flag wasn't carried over, so none of that code was being compiled.
> 
> we are moving the syntax highlighting over to `radicle-interface`. [radicle-dev/radicle-interface#497](https://github.com/radicle-dev/radicle-interface/issues/497) so we won't use that in the http-api, maybe we could remove everything under this feature flag also? wdyt? @cloudhead

Oh sweet! That would be nice to simplify that section of code.

xphoniex:
I ported the routes in `http-api` and the only few places where `Reivision` are used, can simply be replaced with an `Oid`. There's one or two instance, and it's only the `Sha` variant.

FintanH:
> I ported the routes in `http-api` and the only few places where `Reivision` are used, can simply be replaced with an `Oid`. There's one or two instance, and it's only the `Sha` variant.

Sorry, I should have been more clear. I'm referring to [`Revisions`](https://github.com/radicle-dev/radicle-surf/blob/master/source/src/revision.rs#L89-L98), not `Revision`.

xphoniex:
> > I ported the routes in `http-api` and the only few places where `Reivision` are used, can simply be replaced with an `Oid`. There's one or two instance, and it's only the `Sha` variant.
> 
> Sorry, I should have been more clear. I'm referring to [`Revisions`](https://github.com/radicle-dev/radicle-surf/blob/master/source/src/revision.rs#L89-L98), not `Revision`.

Ah, I see. Didn't have to use `Revisions` thus far.

FintanH:
> > > I ported the routes in `http-api` and the only few places where `Reivision` are used, can simply be replaced with an `Oid`. There's one or two instance, and it's only the `Sha` variant.
> > 
> > 
> > Sorry, I should have been more clear. I'm referring to [`Revisions`](https://github.com/radicle-dev/radicle-surf/blob/master/source/src/revision.rs#L89-L98), not `Revision`.
> 
> Ah, I see. Didn't have to use `Revisions` thus far.

Can you perhaps point me to the http-api code that allows you to select a user's fork of the project and their branches?

xphoniex:
> > > > I ported the routes in `http-api` and the only few places where `Reivision` are used, can simply be replaced with an `Oid`. There's one or two instance, and it's only the `Sha` variant.
> > > 
> > > 
> > > Sorry, I should have been more clear. I'm referring to [`Revisions`](https://github.com/radicle-dev/radicle-surf/blob/master/source/src/revision.rs#L89-L98), not `Revision`.
> > 
> > 
> > Ah, I see. Didn't have to use `Revisions` thus far.
> 
> Can you perhaps point me to the http-api code that allows you to select a user's fork of the project and their branches?

I think [`remote_handler`](https://github.com/radicle-dev/heartwood/blob/dave/migrate-api/radicle-httpd/src/api/v1/projects.rs#L246) is what you're describing and perhaps I've implemented it incorrectly?

FintanH:
> I think [`remote_handler`](https://github.com/radicle-dev/heartwood/blob/dave/migrate-api/radicle-httpd/src/api/v1/projects.rs#L246) is what you're describing and perhaps I've implemented it incorrectly?

I'm sure it's fine. I think `Revisions` was first introduced for `radicle-upstream` so there's a good chance it's not used anymore. I think I'll remove it since it does feel strange that there is Heartwood aware code in `radicle-surf`.

cloudhead:
Yup let's remove syntax highlighting code.

Regarding `Revisions` we have our own type in Heartwood called `Remotes` and `Remote`, and it seems pretty similar.

FintanH:
> Yup let's remove syntax highlighting code.
> 
> Regarding `Revisions` we have our own type in Heartwood called `Remotes` and `Remote`, and it seems pretty similar.

Excellent news. I'll remove both :)

keepsimple1:
> radicle-source was folded into radicle-surf from its own crate, the original decision was to place these modules at the top level. This patch series creates a separated hierarchy, source, instead -- see the first commit.

Could you please add doc comments in `src/source.rs` to describe its purpose? For one, I'm always confused ;-)

radicle-surf/src/source.rs
```diff
@@ -0,0 +1,8 @@
+pub mod object;
+pub use object::{blob, tree, Blob, BlobContent, Tree, TreeEntry};
```
keepsimple1, line 2: If we already list `pub use` symbols, normally don't need to set `pub` for the module itself.

radicle-surf/src/source.rs
```diff
@@ -0,0 +1,8 @@
...
+
+pub mod commit;
```
keepsimple1, line 4: same goes for `commit`,  no need to use `pub` as we already list the exported symbols.

radicle-surf/src/source/commit.rs
```diff
@@ -46,7 +45,7 @@ pub struct Stats {
 }
 
 /// Representation of a changeset between two revs.
-#[cfg_attr(feature = "serialize", derive(Serialize))]
+#[cfg_attr(feature = "serde", derive(Serialize))]
 #[derive(Clone)]
 pub struct Commit {
```
keepsimple1, line 50: This seems to be a good opportunity to look into the difference between this `Commit` and `crate::git::commit::Commit`. Even though they are in different modules so the names don't conflict technically, but they require maintainers to dig into the details to figure out. 

For myself, I was always wondering:  why do we have two `commit.rs`? Is it better to merge them, or remove one of them, or to rename one of them?

radicle-surf/src/source.rs
FintanH: I haven't listed all of them :) `Error` is in `object` but since it's a common name I don't want to re-export it here. I'm also fine with leaving it `pub mod` so that if I missed a re-export it would still be available to use. I wouldn't say that's a general rule for me but in this case, I'm not trying to hide anything, just making it easier to use the types re-exported.

radicle-surf/src/source.rs
FintanH: Same goes here

FintanH:
> > radicle-source was folded into radicle-surf from its own crate, the original decision was to place these modules at the top level. This patch series creates a separated hierarchy, source, instead -- see the first commit.
> 
> Could you please add doc comments in `src/source.rs` to describe its purpose? For one, I'm always confused ;-) I think one of the reasons is the name `source` does not click to me, seems vague.

> This seems to be a good opportunity to look into the difference between this `Commit` and `crate::git::commit::Commit`. Even though they are in different modules so the names don't conflict technically, but they require maintainers to dig into the details to figure out.
> 
> For myself, I was always wondering: why do we have two `commit.rs`? Is it better to merge them, or remove one of them, or to rename one of them?

I'll answer both of these here :) If you recall, you [merged `radicle-source` into `radicle-surf`](https://github.com/radicle-dev/radicle-git/commit/58d69ebf7f441cfde043593fc7efbc8912033759) a while back. The purpose of `radicle-source` was to have a layer on top of `radicle-surf` that acted as the explicit "source code browsing" library for the `http-api` (and Radicle Upstream at the time of inception). Hence why the types specifically implement serialization so that the results can be served up the caller of the `http-api`. 

So `source::Commit` has some additional information for being able to display a commit in the browser, for example, this [commit](https://app.radicle.xyz/seeds/clients.radicle.xyz/rad:git:hnrkk9c4zt9thuxhwi1ukxqcrs5tmhbtcsony/remotes/hyn9diwfnytahjq8u3iw63h9jte1ydcatxax3saymwdxqu1zo645pe/commits/be3b830384a9acec16b90a3309c8e72013d1fe24).

I initially thought that we could unify them because I was becoming confused about which one we were talking about, but it was made more clear for me when I put this `Commit` under `source`.

Working on this PR also made me realise that very little of the original `radicle-surf` API is used by `http-api`, it's more focused on using the `source` API. You can check out this PR to see what's being used https://github.com/radicle-dev/heartwood/pull/113

keepsimple1:
> I'll answer both of these here :) If you recall, you [merged `radicle-source` into `radicle-surf`](https://github.com/radicle-dev/radicle-git/commit/58d69ebf7f441cfde043593fc7efbc8912033759) a while back. The purpose of `radicle-source` was to have a layer on top of `radicle-surf` that acted as the explicit "source code browsing" library for the `http-api` (and Radicle Upstream at the time of inception). Hence why the types specifically implement serialization so that the results can be served up the caller of the `http-api`.

Yes I merged `radicle-source` into this crate. Sorry about confusions if any. I wanted to say that, even I merged the `radicle-source`, they still confused me.

Hence I believe a good doc comment would help future users / maintainers to understand what `source` is about. And I also hope maybe we can simplify things further and get rid of some confusions. I will think a bit more on this and discuss with you.

cloudhead:
> @xphoniex @sebastinez May I ask, at the high level, do you have a preference between the `Blob`, `Tree` pair and the `File`, `Directory` pair ? Or do you need to use both pairs?

One thing to consider, if this is mostly a question on naming, is that when working with git, we typically just use tree/blob, eg. even in URL paths in the app, we have eg. `/blob/<oid>` and `/tree/<path>/<oid>` (GitHub does this too). So I think "directory" and "file" might actually be confusing.

keepsimple1:
> One thing to consider, if this is mostly a question on naming, is that when working with git, we typically just use tree/blob, eg. even in URL paths in the app, we have eg. `/blob/<oid>` and `/tree/<path>/<oid>` (GitHub does this too). So I think "directory" and "file" might actually be confusing.

I think it depends on what kind of abstractions we want this crate to provide. For a file system like abstraction, "directory" and "file" make sense to me. But if the users would rather interact with the crate using git terms directly, then "blob" and "tree" might work better for them.

So far I prefer to use "directory" and "file" as I hope this crate can be used by people who are not necessarily familiar with "blob" and "tree". I was one of such people myself a few weeks ago. Even now I probably don't understand some corner cases of "blob" and "tree" usage.  Then again I could have been wrong about our users in this regard.

radicle-surf/src/source.rs
keepsimple1: OK.

cloudhead:
The thing is they are slightly different. A blob is not "just" a file, and a tree is not "just" a directory. They have special properties unique to them. So if we expose those properties (which we should), then I'd call them by their name.

PR #71, opened 2022-11-25
radicle-surf: refactor diff

FintanH:
This is an overhaul of the diff module. The major changes are as follows.

It shouldn't be possible for an added or deleted file to have any other kind line changes other than additions or deletions. So there are now 3 different kinds of changes, Addition, Deletion, or Modification.

Hunk and Hunks now take a generic type parameter to allow for the three types of changes. As a consequence a FileDiff does the same.

For the actual file representations, they fill in their parameters as Addition, Deletion, or Modification.

During this refactor some renaming was also made:
* CreateFile -> Added -- as well as the created field being renamed to added.
* DeleteFile -> Deleted
* MoveFile -> Moved
* CopyFile -> Copied
* ModifiedFile -> Modified

The Stats are also now carried in the Diff instead of being calculated. They can be retrieved from the `git2::Diff`.

The final change was to remove the `source::Commit::stats` field in favour of accessing the `Diff::stats` instead.

The above changes mean that the serialization of `Diff` and `Commit` are not backwards compatible.


radicle-surf/t/src/git/diff.rs
```diff
@@ -171,27 +199,102 @@ fn test_diff_serde() {
...
+@@ -1 +1 @@
+-hello=123
++hello=1234
+"#;
+    let diff = git2::Diff::from_buffer(buf.as_bytes()).unwrap();
+    let diff = Diff::try_from(diff).unwrap();
+    assert_eq!(diff.modified[0].eof, None);
+}
+
+// TODO(xphoniex): uncomment once libgit2 has fixed the bug
```
cloudhead, line 266: Is it fixed by any chance?

FintanH: It doesn't seem to be, @xphoniex what is "the bug" here?

xphoniex: Nope, this is related to https://github.com/radicle-dev/radicle-surf/pull/201, I've followed up once with the author, and I'll try again.

keepsimple1 (approved):
LGTM!

PR #72, opened 2022-11-25
radicle-surf: refactor repo

FintanH:
To have a clear separation of which are public functions and which are private functions, two `impl Repository` blocks are created.

This lead to some refactoring of names and inclusion of helper functions, removing the need for the `git2_repo` function.

The `get_commit_stats` function seemed off, since the `Stats` in this case are to do with the repository. Thus, it does not take a commit anymore and instead uses the `head` commit to calculate the repository's `Stats`.


cloudhead (approved):
LGTM

keepsimple1 (approved):
LGTM!

PR #75, opened 2022-11-30
Enable tests in file_system

keepsimple1:
- Started to update doc for this crate.
- Moved a doc test in a real test case.
- Found one test file `file_system` was not enabled.  Enabled it.
- Fixed a couple bugs after the above test was enabled.

### DIFF
--- a/radicle-surf/examples/browsing.rs
+++ b/radicle-surf/examples/browsing.rs
@@ -51,10 +51,10 @@ fn main() {
 fn print_directory(d: &Directory, repo: &Repository, indent_level: usize) {
     let indent = " ".repeat(indent_level * 4);
     println!("{}{}/", &indent, d.name());
-    for entry in d.entries(repo).unwrap().entries() {
+    for entry in d.entries(repo).unwrap() {
         match entry {
             directory::Entry::File(f) => println!("    {}{}", &indent, f.name()),
-            directory::Entry::Directory(d) => print_directory(d, repo, indent_level + 1),
+            directory::Entry::Directory(d) => print_directory(&d, repo, indent_level + 1),
         }
     }
 }
--- a/radicle-surf/src/file_system/directory.rs
+++ b/radicle-surf/src/file_system/directory.rs
@@ -44,6 +44,8 @@ pub mod error {
         Entry(#[from] Entry),
         #[error(transparent)]
         File(#[from] File),
+        #[error("the path {0} is not valid")]
+        InvalidPath(String),
     }
 
     #[derive(Debug, Error, PartialEq, Eq)]
@@ -178,6 +180,19 @@ impl Entries {
     }
 }
 
+impl Iterator for Entries {
+    type Item = Entry;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // Can be improved when `pop_first()` is stable for BTreeMap.
+        let next_key = match self.listing.keys().next() {
+            Some(k) => k.clone(),
+            None => return None,
+        };
+        self.listing.remove(&next_key)
+    }
+}
+
 /// An `Entry` is either a [`File`] entry or a [`Directory`] entry.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum Entry {
@@ -226,6 +241,16 @@ impl Entry {
             Entry::File(file) => file.location(),
             Entry::Directory(directory) => directory.location(),
         }
+    }
+
+    /// Returns `true` if the `Entry` is a file.
+    pub fn is_file(&self) -> bool {
+        matches!(self, Entry::File(_))
+    }
+
+    /// Returns `true` if the `Entry` is a directory.
+    pub fn is_directory(&self) -> bool {
+        matches!(self, Entry::Directory(_))
     }
 
     pub(crate) fn from_entry(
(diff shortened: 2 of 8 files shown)


radicle-surf/t/src/file_system.rs
```diff
@@ -50,6 +49,49 @@ mod directory {
...
+        assert!(root_contents[6].is_directory());
+        assert_eq!(root_contents[3].name(), "bin");
+        assert_eq!(root_contents[4].name(), "src");
+        assert_eq!(root_contents[5].name(), "text");
+        assert_eq!(root_contents[6].name(), "this");
+
+        let src = root
+            .find_directory(Path::new("src"), &repo)
+            .unwrap()
+            .unwrap();
```
keepsimple1, line 87: I didn't like the double unwrap. I think it is probably a better interface if `find_directory` returns `Result<>`, not `Result<Option<>>`. 

If this directory was not found, we can return an error like `NotFound`.

radicle-surf/examples/browsing.rs
```diff
@@ -51,10 +51,10 @@ fn main() {
 fn print_directory(d: &Directory, repo: &Repository, indent_level: usize) {
     let indent = " ".repeat(indent_level * 4);
     println!("{}{}/", &indent, d.name());
-    for entry in d.entries(repo).unwrap().entries() {
+    for entry in d.entries(repo).unwrap() {
```
keepsimple1, line 54: Double `entries()` in the existing code seems confusing, so I added `impl Iterator` for  `Entries` to support this simpler usage. 

Another option is to rename the first `entries()` method, maybe.

radicle-surf/src/file_system/directory.rs
```diff
@@ -327,31 +350,31 @@ impl Directory {
...
+            .get_path(path)
             .map(Some)
             .or_matches::<git2::Error, _, _>(is_not_found_err, || Ok(None))?;
+        let parent = path
+            .parent()
+            .ok_or_else(|| git::Error::PathNotFound(path.to_path_buf()))?;
 
         Ok(entry
-            .and_then(|entry| Entry::from_entry(&entry, path.as_ref().to_path_buf()).transpose())
+            .and_then(|entry| Entry::from_entry(&entry, parent.to_path_buf()).transpose())
```
keepsimple1, line 371: This was what apparently failed the tests.

radicle-surf/src/file_system/directory.rs
```diff
@@ -327,31 +350,31 @@ impl Directory {
     /// Find the [`Entry`] found at `path`, if it exists.
     pub fn find_entry<P>(
         &self,
-        path: &P,
+        path: P,
```
FintanH, line 353: Why did you change taking by reference here?

radicle-surf/t/src/file_system.rs
FintanH: I'm on the fence about having a `NotFound` error for a `find_*` function. I generally prefer that the caller has the option to handle that case explicitly through the `Option`, or they use their own `NotFound` error and `ok_or`. Another option is to have a `try_find_file`, but that still means we end up hiding a `NotFound` in our error variants.

The fact that this is a double unwrap is only because we're in a testing environment.

FintanH (dismissed):
Nice, thanks for catching the bug! Left some comments below :)

radicle-surf/t/src/file_system.rs
keepsimple1: I will leave it as is for now.

radicle-surf/src/file_system/directory.rs
keepsimple1: Because `P` is already a reference: `P: AsRef<Path>`.  And in practice `Path::new()` returns `&Path`.

keepsimple1: Because `P` is already a reference: `P: AsRef<Path>`.  And in practice `Path::new()` returns `&Path`.

keepsimple1: It is also consistent with the existing `find_directory`

FintanH: > Because `P` is already a reference: `P: AsRef<Path>`. And in practice `Path::new()` returns `&Path`.

That's not necessarily what `P: AsRef<Path>` means. If `P: AsRef<Path>` that implies the following code:
```rust
 // for some P
impl AsRef<Path> for P {
  fn as_ref(&self) -> &Path {
    /* code */
  }
}
```

The `&self` means that we have a reference of type `P`. So when we take ownership of `P` in the way you're doing above that satisfies the borrow checker since owned implies borrowed. That's quite restrictive. Say if I wanted to use the same `path` for multiple functions including `find_entry`, I would have to `clone` to do so -- when the `clone` is not really needed.

I must have missed this with `find_directory` so I would say we change `find_entry` back and fix `find_directory` :)

keepsimple1: >  Say if I wanted to use the same path for multiple functions including find_entry, I would have to clone to do so -- when the clone is not really needed.

In practice, `Path::new()` returns `&Path`, and you probably always have `&Path` not `Path` and no need to clone `Path`.

FintanH: > > Say if I wanted to use the same path for multiple functions including find_entry, I would have to clone to do so -- when the clone is not really needed.
> 
> In practice, [`Path::new()`](https://doc.rust-lang.org/std/path/struct.Path.html#method.new) returns `&Path`, and you probably always have `&Path` not `Path` and no need to clone `Path`.
> 
> I will make the change.

Right, so even more reason to have a borrowed parameter :smile: You would have to use `to_path_buf` to make it owned :)

FintanH (approved):

PR #79, opened 2022-12-06
refactor Tree, Blob, Directory and File

keepsimple1:
This is motivated by my [attempt](https://github.com/radicle-dev/heartwood/pull/149) to use the new `radicle-surf` API in `heartwood/radicle-httpd`.  The main changes are:

1. Refactored `Tree` and `Blob` so that they represent the "contents" without embedding the path or the file name. In contrast, `Directory` and `File` include their own `path` and `name`.
2. Provided `repo.tree()`, `repo.blob()`, `repo.directory()` and `repo.file()` methods. 
3. Provided `directory.last_commit()` to show that `Directory` can provide everything a user might wanted from a `Tree`, although in a more "lazy" style.

The immediate goal is to allow the user to try out the `tree` and `blob` methods since it seems they feel familiar with them, and less confused comparing with `Directory` and `File`. Once people start to use them, we can understand better and improve / add / delete them.

### DIFF
--- a/radicle-surf/src/file_system/directory.rs
+++ b/radicle-surf/src/file_system/directory.rs
@@ -31,7 +31,7 @@ use git2::Blob;
 use radicle_git_ext::{is_not_found_err, Oid};
 use radicle_std_ext::result::ResultExt as _;
 
-use crate::git::{Repository, Revision};
+use crate::git::{Commit, Repository, Revision};
 
 pub mod error {
     use thiserror::Error;
@@ -79,6 +79,8 @@ pub struct File {
     prefix: PathBuf,
     /// The object identifier of the git blob of this file.
     id: Oid,
+    /// The commit that created this file version.
+    last_commit: Commit,
 }
 
 impl File {
@@ -88,14 +90,19 @@ impl File {
     /// so should not end in `name`.
     ///
     /// The `id` must point to a git blob.
-    pub(crate) fn new(name: String, prefix: PathBuf, id: Oid) -> Self {
+    pub(crate) fn new(name: String, prefix: PathBuf, id: Oid, last_commit: Commit) -> Self {
         debug_assert!(
             !prefix.ends_with(&name),
             "prefix = {:?}, name = {}",
             prefix,
             name
         );
-        Self { name, prefix, id }
+        Self {
+            name,
+            prefix,
+            id,
+            last_commit,
+        }
     }
 
     /// The name of this `File`.
@@ -256,12 +263,26 @@ impl Entry {
     pub(crate) fn from_entry(
         entry: &git2::TreeEntry,
         path: PathBuf,
+        repo: &Repository,
+        parent_commit: Oid,
     ) -> Result<Option<Self>, error::Entry> {
         let name = entry.name().ok_or(error::Entry::Utf8Error)?.to_string();
         let id = entry.id().into();
+        let escaped_name = name.replace('\\', r"\\");
+        let entry_path = path.join(escaped_name);
+        // FIXME: I don't like to use FIXME, but here it is. I would
+        // like to simplify the error definitions and then fix these
+        // unwrap(s).
+        let last_commit = repo
+            .last_commit(&entry_path, parent_commit)
+            .unwrap()
+            .unwrap();
+
         Ok(entry.kind().and_then(|kind| match kind {
-            git2::ObjectType::Tree => Some(Self::Directory(Directory::new(name, path, id))),
-            git2::ObjectType::Blob => Some(Self::File(File::new(name, path, id))),
+            git2::ObjectType::Tree => {
+                Some(Self::Directory(Directory::new(name, path, id, last_commit)))
+            },
+            git2::ObjectType::Blob => Some(Self::File(File::new(name, path, id, last_commit))),
             _ => None,
         }))
     }
@@ -285,16 +306,18 @@ pub struct Directory {
     prefix: PathBuf,
     /// The object identifier of the git tree of this directory.
     id: Oid,
+    /// The commit that created this directory version.
+    last_commit: Commit,
 }
 
+const ROOT_DIR: &str = "";
+
 impl Directory {
-    /// Creates a directory given its `id`.
+    /// Creates a directory given its `tree_id`.
     ///
     /// The `name` and `prefix` are both set to be empty.
-    ///
-    /// The `id` must point to a `git` tree.
-    pub(crate) fn root(id: Oid) -> Self {
-        Self::new("".to_string(), PathBuf::new(), id)
+    pub(crate) fn root(tree_id: Oid, repo_commit: Commit) -> Self {
+        Self::new(ROOT_DIR.to_string(), PathBuf::new(), tree_id, repo_commit)
     }
 
     /// Creates a directory given its `name` and `id`.
@@ -303,14 +326,19 @@ impl Directory {
     /// so should not end in `name`.
     ///
     /// The `id` must point to a `git` tree.
-    pub(crate) fn new(name: String, prefix: PathBuf, id: Oid) -> Self {
+    pub(crate) fn new(name: String, prefix: PathBuf, id: Oid, last_commit: Commit) -> Self {
         debug_assert!(
             name.is_empty() || !prefix.ends_with(&name),
             "prefix = {:?}, name = {}",
             prefix,
             name
         );
-        Self { name, prefix, id }
+        Self {
+            name,
+            prefix,
+            id,
+            last_commit,
+        }
     }
 
     /// Get the name of the current `Directory`.
@@ -354,9 +382,10 @@ impl Directory {
         let mut error = None;
         let path = self.path();
 
-        // Walks only the first level of entries.
-        tree.walk(git2::TreeWalkMode::PreOrder, |_, entry| {
-            match Entry::from_entry(entry, path.clone()) {
+        // Walks only the first level of entries. And `_entry_path` is always
+        // empty for the first level.
+        tree.walk(git2::TreeWalkMode::PreOrder, |_entry_path, entry| {
+            match Entry::from_entry(entry, path.clone(), repo, self.last_commit.id) {
                 Ok(Some(entry)) => match entry {
                     Entry::File(_) => {
                         entries.insert(entry.name().clone(), entry);
@@ -382,6 +411,11 @@ impl Directory {
         }
     }
 
+    /// Returns the last commit that created or modified this directory.
+    pub fn last_commit(&self) -> &Commit {
+        &self.last_commit
+    }
+
     /// Find the [`Entry`] found at `path`, if it exists.
     pub fn find_entry<P>(
         &self,
@@ -401,9 +435,13 @@ impl Directory {
         let parent = path
             .parent()
             .ok_or_else(|| error::Directory::InvalidPath(path.to_string_lossy().to_string()))?;
+        let root_path = self.path().join(parent);
 
         Ok(entry
-            .and_then(|entry| Entry::from_entry(&entry, parent.to_path_buf()).transpose())
+            .and_then(|entry| {
+                Entry::from_entry(&entry, root_path.to_path_buf(), repo, self.last_commit.id)
+                    .transpose()
+            })
             .transpose()
             .unwrap())
     }
@@ -424,6 +462,8 @@ impl Directory {
     }
 
     /// Find the `Directory` found at `path`, if it exists.
+    ///
+    /// If `path` is `ROOT_DIR` (i.e. an empty path), returns self.
     pub fn find_directory<P>(
         &self,
         path: &P,
@@ -432,6 +472,10 @@ impl Directory {
     where
         P: AsRef<Path>,
     {
+        if path.as_ref() == Path::new(ROOT_DIR) {
+            return Ok(Some(self.clone()));
+        }
+
         Ok(match self.find_entry(path, repo)? {
             Some(Entry::Directory(d)) => Some(d),
             _ => None,
--- a/radicle-surf/src/git/repo.rs
+++ b/radicle-surf/src/git/repo.rs
@@ -28,7 +28,10 @@ use thiserror::Error;
 
 use crate::{
     diff::{self, *},
-    file_system::{directory::FileContent, Directory},
+    file_system::{
+        directory::{self, File, FileContent},
+        Directory,
+    },
     git::{
         commit,
         glob,
@@ -44,6 +47,7 @@ use crate::{
         Tag,
         ToCommit,
     },
+    source::{self, commit::Header, Blob, Tree, TreeEntry},
 };
 
 pub mod iter;
@@ -65,6 +69,10 @@ pub enum Error {
     #[error(transparent)]
     Diff(#[from] diff::git::error::Diff),
     /// A wrapper around the generic [`git2::Error`].
+    #[error(transparent)]
+    Directory(#[from] directory::error::Directory),
+    #[error(transparent)]
+    File(#[from] directory::error::File),
     #[error(transparent)]
     Git(#[from] git2::Error),
     #[error(transparent)]
@@ -249,12 +257,72 @@ impl Repository {
             .map_err(|err| Error::ToCommit(err.into()))?;
         let git2_commit = self.inner.find_commit((commit.id).into())?;
         let tree = git2_commit.as_object().peel_to_tree()?;
-        Ok(Directory::root(tree.id().into()))
+        Ok(Directory::root(tree.id().into(), commit))
+    }
+
+    /// Returns a [`Directory`] for `path` in `commit`.
+    pub fn directory<C: ToCommit, P: AsRef<Path>>(
+        &self,
+        commit: C,
+        path: &P,
+    ) -> Result<Directory, Error> {
+        let root = self.root_dir(commit)?;
+        root.find_directory(path, self)?
+            .ok_or_else(|| Error::PathNotFound(path.as_ref().to_path_buf()))
+    }
+
+    /// Returns a [`File`] for `path` in `commit`.
+    pub fn file<C: ToCommit, P: AsRef<Path>>(&self, commit: C, path: &P) -> Result<File, Error> {
+        let root = self.root_dir(commit)?;
+        root.find_file(path, self)?
+            .ok_or_else(|| Error::PathNotFound(path.as_ref().to_path_buf()))
+    }
+
+    /// Returns a [`Tree`] for `path` in `commit`.
+    pub fn tree<C: ToCommit, P: AsRef<Path>>(&self, commit: C, path: &P) -> Result<Tree, Error> {
+        let commit = commit
+            .to_commit(self)
+            .map_err(|e| Error::ToCommit(e.into()))?;
+        let dir = self.directory(commit.id, path)?;
+        let mut entries = dir
+            .entries(self)?
+            .map(|en| {
+                let name = en.name().to_string();
+                let path = en.path();
+                let commit = self
+                    .last_commit(&path, commit.id)?
+                    .ok_or(Error::PathNotFound(path))?;
+                let commit_header = Header::from(commit);
+                Ok(TreeEntry::new(name, en.into(), commit_header))
+            })
+            .collect::<Result<Vec<TreeEntry>, Error>>()?;
+        entries.sort();
+
+        let last_commit = self
+            .last_commit(path, commit)?
+            .ok_or_else(|| Error::PathNotFound(path.as_ref().to_path_buf()))?;
+        let header = source::commit::Header::from(last_commit);
+        Ok(Tree::new(dir.id(), entries, header))
+    }
+
+    /// Returns a [`Blob`] for `path` in `commit`.
+    pub fn blob<C: ToCommit, P: AsRef<Path>>(&self, commit: C, path: &P) -> Result<Blob, Error> {
+        let commit = commit
+            .to_commit(self)
+            .map_err(|e| Error::ToCommit(e.into()))?;
+        let file = self.file(commit.id, path)?;
+        let last_commit = self
+            .last_commit(path, commit)?
+            .ok_or_else(|| Error::PathNotFound(path.as_ref().to_path_buf()))?;
+        let header = source::commit::Header::from(last_commit);
+
+        let content = file.content(self)?;
+        Ok(Blob::new(file.id(), content.as_bytes(), header))
     }
 
     /// Returns the last commit, if exists, for a `path` in the history of
     /// `rev`.
-    pub fn last_commit<P, C>(&self, path: P, rev: C) -> Result<Option<Commit>, Error>
+    pub fn last_commit<P, C>(&self, path: &P, rev: C) -> Result<Option<Commit>, Error>
     where
         P: AsRef<Path>,
         C: ToCommit,
@@ -444,7 +512,7 @@ impl Repository {
 
         let mut opts = git2::DiffOptions::new();
         if let Some(path) = path {
-            opts.pathspec(path);
+            opts.pathspec(path.to_string_lossy().to_string());
             // We're skipping the binary pass because we won't be inspecting deltas.
             opts.skip_binary_check(true);
         }
--- a/radicle-surf/src/source/object/blob.rs
+++ b/radicle-surf/src/source/object/blob.rs
@@ -18,76 +18,52 @@
 //! Represents git object type 'blob', i.e. actual file contents.
 //! See git [doc](https://git-scm.com/book/en/v2/Git-Internals-Git-Objects) for more details.
 
-use std::{path::Path, str};
+use std::str;
 
+use radicle_git_ext::Oid;
 #[cfg(feature = "serde")]
 use serde::{
     ser::{SerializeStruct as _, Serializer},
     Serialize,
 };
 
-use crate::{
-    file_system::{File, FileContent},
-    git::{self, Repository},
-    source::{commit, object::Error},
-};
+use crate::source::commit;
 
-/// File data abstraction.
+/// Represents a git blob object.
 pub struct Blob {
-    pub file: File,
-    pub content: BlobContent,
-    pub commit: Option<commit::Header>,
+    id: Oid,
+    content: BlobContent,
+    commit: commit::Header,
 }
 
 impl Blob {
     /// Returns the [`Blob`] for a file at `revision` under `path`.
-    ///
-    /// # Errors
-    ///
-    /// Will return [`Error`] if the project doesn't exist or a surf interaction
-    /// fails.
-    pub fn new<P, R>(repo: &Repository, revision: &R, path: &P) -> Result<Blob, Error>
-    where
-        P: AsRef<Path>,
-        R: git::Revision,
-    {
-        Self::make_blob(repo, revision, path, |c| BlobContent::from(c))
-    }
-
-    fn make_blob<P, R>(
-        repo: &Repository,
-        revision: &R,
-        path: &P,
-        content: impl FnOnce(FileContent) -> BlobContent,
-    ) -> Result<Blob, Error>
-    where
-        P: AsRef<Path>,
-        R: git::Revision,
-    {
-        let path = path.as_ref();
-        let root = repo.root_dir(revision)?;
-
-        let file = root
-            .find_file(&path, repo)?
-            .ok_or_else(|| Error::PathNotFound(path.to_path_buf()))?;
-
-        let last_commit = repo
-            .last_commit(path, revision)?
-            .map(|c| commit::Header::from(&c));
-
-        let content = content(file.content(repo)?);
-
-        Ok(Blob {
-            file,
+    pub(crate) fn new(id: Oid, content: &[u8], commit: commit::Header) -> Self {
+        let content = BlobContent::from(content);
+        Self {
+            id,
             content,
-            commit: last_commit,
-        })
+            commit,
+        }
     }
 
     /// Indicates if the content of the [`Blob`] is binary.
     #[must_use]
     pub fn is_binary(&self) -> bool {
         matches!(self.content, BlobContent::Binary(_))
+    }
+
+    pub fn object_id(&self) -> Oid {
+        self.id
+    }
+
+    pub fn content(&self) -> &BlobContent {
+        &self.content
+    }
+
+    /// Returns the commit that created this blob.
+    pub fn commit(&self) -> &commit::Header {
+        &self.commit
     }
 }
 
@@ -102,8 +78,6 @@ impl Serialize for Blob {
         state.serialize_field("binary", &self.is_binary())?;
         state.serialize_field("content", &self.content)?;
         state.serialize_field("lastCommit", &self.commit)?;
-        state.serialize_field("name", &self.file.name())?;
-        state.serialize_field("path", &self.file.path())?;
         state.end()
     }
 }
@@ -115,6 +89,24 @@ pub enum BlobContent {
     Plain(String),
     /// Content is binary and needs special treatment.
     Binary(Vec<u8>),
+}
+
+impl BlobContent {
+    /// Returns the size of this `BlobContent`.
+    pub fn size(&self) -> usize {
+        match self {
+            Self::Plain(content) => content.len(),
+            Self::Binary(bytes) => bytes.len(),
+        }
+    }
+
+    /// Returns the content as bytes.
+    pub fn as_bytes(&self) -> &[u8] {
+        match self {
+            Self::Plain(content) => content.as_bytes(),
+            Self::Binary(bytes) => &bytes[..],
+        }
+    }
 }
 
 #[cfg(feature = "serde")]
@@ -133,12 +125,11 @@ impl Serialize for BlobContent {
     }
 }
 
-impl From<FileContent<'_>> for BlobContent {
-    fn from(content: FileContent) -> Self {
-        let content = content.as_bytes();
-        match str::from_utf8(content) {
+impl From<&[u8]> for BlobContent {
+    fn from(bytes: &[u8]) -> Self {
+        match str::from_utf8(bytes) {
             Ok(utf8) => BlobContent::Plain(utf8.to_owned()),
-            Err(_) => BlobContent::Binary(content.to_owned()),
+            Err(_) => BlobContent::Binary(bytes.to_owned()),
         }
     }
 }
--- a/radicle-surf/src/source/object/tree.rs
+++ b/radicle-surf/src/source/object/tree.rs
@@ -18,125 +18,213 @@
 //! Represents git object type 'tree', i.e. like directory entries in Unix.
 //! See git [doc](https://git-scm.com/book/en/v2/Git-Internals-Git-Objects) for more details.
 
-use std::path::Path;
+use std::cmp::Ordering;
 
+use radicle_git_ext::Oid;
 #[cfg(feature = "serde")]
 use serde::{
     ser::{SerializeStruct as _, Serializer},
     Serialize,
 };
 
-use crate::{
-    file_system::{directory, Directory},
-    git::{self, Repository},
-    source::{commit, object::Error},
-};
+use crate::{file_system::directory, source::commit};
 
-/// Result of a directory listing, carries other trees and blobs.
+/// Represents a tree object as in git. It is essentially the content of
+/// one directory. Note that multiple directories can have the same content,
+/// i.e. have the same tree object. Hence this struct does not embed its path.
 #[derive(Clone, Debug)]
 pub struct Tree {
-    pub directory: Directory,
-    pub commit: Option<commit::Header>,
-    /// Entries listed in that tree result.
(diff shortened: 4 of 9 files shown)


radicle-surf/src/git/repo.rs
```diff
@@ -252,6 +257,28 @@ impl Repository {
         Ok(Directory::root(tree.id().into()))
     }
 
+    /// Returns a [Directory] for `path` in `commit`.
+    pub fn directory<C: ToCommit>(&self, commit: C, path: &str) -> Result<Directory, Error> {
```
FintanH, line 261: I would change the `path` parameter to be more general in this case:
```suggestion
    pub fn directory<C: ToCommit, P: AsRef<Path>>(&self, commit: C, path: &P) -> Result<Directory, Error> {
```

radicle-surf/src/git/repo.rs
```diff
@@ -252,6 +257,28 @@ impl Repository {
         Ok(Directory::root(tree.id().into()))
     }
 
+    /// Returns a [Directory] for `path` in `commit`.
```
FintanH, line 260:
```suggestion
    /// Returns a [`Directory`] for `path` in `commit`.
```

radicle-surf/src/git/repo.rs
```diff
@@ -252,6 +257,28 @@ impl Repository {
...
+        let path = std::path::Path::new(path);
+        let root = self.root_dir(commit)?;
+        let sub_dir = root.find_directory(&path, self)?;
+        match sub_dir {
+            Some(d) => Ok(d),
+            None => Err(Error::PathNotFound(path.to_path_buf())),
+        }
+    }
+
+    /// Returns a [File] for `path` in `commit`.
```
FintanH, line 271:
```suggestion
    /// Returns a [`File`] for `path` in `commit`.
```

radicle-surf/src/git/repo.rs
```diff
@@ -252,6 +257,28 @@ impl Repository {
...
 
+    /// Returns a [Directory] for `path` in `commit`.
+    pub fn directory<C: ToCommit>(&self, commit: C, path: &str) -> Result<Directory, Error> {
+        let path = std::path::Path::new(path);
+        let root = self.root_dir(commit)?;
+        let sub_dir = root.find_directory(&path, self)?;
+        match sub_dir {
+            Some(d) => Ok(d),
+            None => Err(Error::PathNotFound(path.to_path_buf())),
+        }
```
FintanH, line 268:
```suggestion
        root.find_file(&path, self)?.ok_or_else(|| Error::PathNotFound(path.to_path_buf()))
```

radicle-surf/src/git/repo.rs
```diff
@@ -252,6 +257,28 @@ impl Repository {
...
+
+    /// Returns a [File] for `path` in `commit`.
+    pub fn file<C: ToCommit>(&self, commit: C, path: &str) -> Result<File, Error> {
+        let path = std::path::Path::new(path);
+        let root = self.root_dir(commit)?;
+        let f = root.find_file(&path, self)?;
+        match f {
+            Some(file) => Ok(file),
+            None => Err(Error::PathNotFound(path.to_path_buf())),
+        }
```
FintanH, line 279:
```suggestion
        root.find_file(&path, self)?.ok_or_else(|| Error::PathNotFound(path.to_path_buf()))
```

radicle-surf/src/git/repo.rs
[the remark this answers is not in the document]
FintanH: > I considered if we should return `Result<Option<Directory, Error>>` here. With all things considered, especially how the caller uses this API, I think this is a good trade-off: if the `path` is not found, we will return `Error::PathNotFound`. It will be much more ergonomic to use.

Happy with this decision :)

> And, I would also recommend we change the [existing `Directory::find_directory`](https://github.com/radicle-dev/radicle-git/blob/1316a13dfc82ed79117aa0c17de9ad703c9bb180/radicle-surf/src/file_system/directory.rs#L430) to return `Result<>` instead of `Result<Option<>>`.

I believe we discussed this already. If we're going to have a fallible version then I'd suggest we add a `try_find_directory` and `try_find_file`, and keep the `Option` functions.

radicle-surf/src/file_system/directory.rs
```diff
@@ -277,6 +279,7 @@ impl Entry {
 ///
 /// [git-tree]: https://git-scm.com/book/en/v2/Git-Internals-Git-Objects
 #[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
```
FintanH, line 282: I would note that the `source::Tree` serialization had a specific structure which wouldn't be matched by this `Serialize`. Can you shed some light on why you added this?

FintanH (commented):
These additions make sense. Have some suggestions and questions :)

radicle-surf/src/git/repo.rs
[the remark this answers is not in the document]
keepsimple1: > I believe we discussed this already. If we're going to have a fallible version then I'd suggest we add a `try_find_directory` and `try_find_file`, and keep the `Option` functions.

It is [fallible](https://github.com/radicle-dev/radicle-git/blob/1316a13dfc82ed79117aa0c17de9ad703c9bb180/radicle-surf/src/file_system/directory.rs#L430) already.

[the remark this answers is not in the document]
FintanH: _More_ fallible then :) It's not a hill I'll die on, all the same. If you want to add a `NotFound` variant to the error types and not return `Option` then that's fine :+1:

radicle-surf/src/file_system/directory.rs
keepsimple1: It is added to support [this](https://github.com/radicle-dev/heartwood/pull/149/files#diff-85a73839d407aa508b9e78b3756b0e9b6e8d1087a0cafac6a8325a8cf427b65eR238).

Did you mean you would like to use `radicle_surf::source::Tree` serialization instead in `heartwood::radicle_httpd` ? 

Btw, I really would like to sort out the duo of `Directory` / `Tree` ,  and `File` / `Blob`.  As we discussed before, I think it's better to stick with only one pair, i.e. either (`Directory`, `File`) or (`Tree` , `Blob`) .   Personally I like to have `Directory`, `File`,  but I'm also OK if we stick with `Tree`, `Blob` and completely get rid of `Directory` , `File`. 

When two things are 90% same but 10% different, that's really confusing.

FintanH: Ya, so as I pointed out in my comment here https://github.com/radicle-dev/heartwood/pull/149/files#r1043374100, this has a completely different serialization to what `httpd` was using before -- which will be no good for when they run the web app and suddenly there's information missing.

First off, do you understand why there are currently two different representations? I've tried to explain in a few ways so I'd like to understand how much of that you're picking up and if I need to rephrase or give a longer explanation. Would be good to get you to tell me what you've gathered so far.

keepsimple1: Well, my understanding is that the only thing is `Tree` that is no in `Directory` is :
```
    pub commit: Option<commit::Header>,
```
which is the `last_commit` of the `directory`.  We can add `last_commit` into `Directory` if needed.

>  if I need to rephrase or give a longer explanation.

Instead of a longer explanation, I'm wondering, what is the No.1 reason for having `Tree` in addition to `Directory` in `radicle-surf`'s public API?

radicle-surf/src/git/repo.rs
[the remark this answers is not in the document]
keepsimple1: As this change would not be in the scope of this PR, I've opened an issue to track it: https://github.com/radicle-dev/radicle-git/issues/82

radicle-surf/src/file_system/directory.rs
FintanH: https://github.com/radicle-dev/radicle-git/blob/main/radicle-surf/src/source/object/tree.rs#L36-L41
```rust
pub struct Tree {
    pub directory: Directory,
    pub commit: Option<commit::Header>,
    /// Entries listed in that tree result.
    pub entries: Vec<TreeEntry>,
}
```
it also holds the entries where `Directory` has the `Oid`. This is necessary because the httpd API wants to list the directories entries. 

> Instead of a longer explanation, I'm wondering, what is the No.1 reason for having Tree in addition to Directory in radicle-surf's public API?

So, the reason we have `Tree` is for httpd -- which is our one and only user of surf right now. It requires the extra information of a last commit header and the tree's entries (in our current design of `Directory` it doesn't have this as mentioned above). Same goes for `Blob`. 

So the No.1 reason, in short, is to support httpd usage.

Is it necessary to have two separate types, well these are the scenarios I foresee.
1. We combine `Directory` and `Tree` into a single type, let's say `Tree`. Now we have to calculate all the necessary information up front, i.e. last commits and entries.
2. We keep the two separate types.
3. We find a way for `Directory` to be extensible, i.e. there's a way to add information to it that can be serialized.

keepsimple1: > it also holds the entries where Directory has the Oid. This is necessary because the httpd API wants to list the directories entries.

Wouldn't [`Directory::entries()`](it also holds the entries where Directory has the Oid. This is necessary because the httpd API wants to list the directories entries.) method do the job?

radicle-surf/t/src/lib.rs
```diff
@@ -6,3 +6,6 @@ mod git;
 
 #[cfg(test)]
 mod file_system;
+
+#[cfg(test)]
+mod object;
```
FintanH, line 11: Can we do `source/object.rs` because we'll likely test other things in `source` too. Or `source/tree.rs`

radicle-surf/t/src/object.rs
```diff
@@ -0,0 +1,27 @@
...
+    );
+
+    let tree_oid = tree.object_id();
+    assert_eq!(
+        tree_oid.to_string(),
+        "dbd5d80c64a00969f521b96401a315e9481e9561"
+    );
+
+    let json = serde_json::json!(tree).to_string();
+    println!("{}", &json);
```
FintanH, line 26: Can we assert the output instead of printing.

radicle-surf/src/source/object/tree.rs
```diff
@@ -126,16 +128,22 @@ impl Serialize for TreeEntry {
...
             "kind",
             match self.entry {
-                directory::Entry::File(_) => "blob",
-                directory::Entry::Directory(_) => "directory",
+                Entry::Blob(_) => "blob",
+                Entry::Tree(_) => "tree",
             },
         )?;
+        state.serialize_field(
+            "oid",
```
FintanH, line 140: I like the addition of the `oid` :+1:

radicle-surf/src/source/object/tree.rs
```diff
@@ -126,16 +128,22 @@ impl Serialize for TreeEntry {
     {
         const FIELDS: usize = 4;
         let mut state = serializer.serialize_struct("TreeEntry", FIELDS)?;
-        state.serialize_field("path", &self.entry.location())?;
-        state.serialize_field("name", &self.entry.name())?;
-        state.serialize_field("lastCommit", &None::<commit::Header>)?;
```
FintanH, line 131: Again, why are you choosing to remove these fields?

radicle-surf/src/source/object/tree.rs
```diff
@@ -20,24 +20,25 @@
...
 };
 
 /// Result of a directory listing, carries other trees and blobs.
 pub struct Tree {
-    pub directory: Directory,
-    pub commit: Option<commit::Header>,
+    id: Oid,
     /// Entries listed in that tree result.
     pub entries: Vec<TreeEntry>,
+    pub commit: commit::Header,
```
FintanH, line 41: Why did you make this decision of removing the `Option`?

radicle-surf/src/source/object/blob.rs
```diff
@@ -101,9 +97,7 @@ impl Serialize for Blob {
         let mut state = serializer.serialize_struct("Blob", FIELDS)?;
         state.serialize_field("binary", &self.is_binary())?;
         state.serialize_field("content", &self.content)?;
-        state.serialize_field("lastCommit", &self.commit)?;
-        state.serialize_field("name", &self.file.name())?;
-        state.serialize_field("path", &self.file.location())?;
```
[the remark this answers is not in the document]
keepsimple1: This is not final change for `Blob`, sorry. I will update this part.

radicle-surf/src/source/object/tree.rs
```diff
@@ -92,29 +73,50 @@ impl Serialize for Tree {
...
         state.serialize_field("entries", &self.entries)?;
         state.serialize_field("lastCommit", &self.commit)?;
-        state.serialize_field("name", &self.directory.name())?;
-        state.serialize_field("path", &self.directory.location())?;
         state.end()
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub enum Entry {
```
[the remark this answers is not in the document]
keepsimple1: Like I mentioned in https://github.com/radicle-dev/heartwood/pull/149#discussion_r1045467640 , I am trying to remove `Directory` from `Tree`, and that includes `TreeEntry` as well. 

In other words, the idea is that `Tree` can (and should) be self-contained without relying on `Directory`.

radicle-surf/t/src/lib.rs
keepsimple1: I can do that for this PR. Since we're here, I have been thinking of removing the `source/` and put `object.rs` and `object/` directly in `radicle-surf/src`.

radicle-surf/t/src/object.rs
keepsimple1: will do.

radicle-surf/src/source/object/tree.rs
keepsimple1: thanks!

radicle-surf/src/source/object/tree.rs
keepsimple1: `name` is kept.  Now it's from `&self.name`,  not `&self.entry.name()`. 

`path` is removed for two reasons: 
1. It's not part of the entry, in the sense of entry in git tree object.
2. The caller that generates the tree already knows the path. In fact, the path is one of the input parameters.

`lastCommit` is removed as it is always `None`.  We provide a new method to get the last commit for a `TreeEntry`. If needed, we could add the last commit info in the serialization.

radicle-surf/src/source/object/tree.rs
keepsimple1: For any tree, there must be a commit that created the tree object in the first place, is the correct? I.e. this commit should never be `None`.

radicle-surf/src/source/object/tree.rs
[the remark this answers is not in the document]
FintanH: Ok, I'll wait and see how it turns out. I imagine it's mostly copying over the data from `Directory`/`Entry`

radicle-surf/t/src/lib.rs
FintanH: And the `person.rs` also goes up into `src`?

radicle-surf/src/source/object/tree.rs
FintanH: Ya, I believe so :) Just wanted to make sure it was for the right reason :grin: Thanks!

radicle-surf/t/src/lib.rs
keepsimple1: The only thing in `person.rs` is `struct Person`, and it is only used in `source/commit.rs`. I propose to move `struct Person` into `source/commit.rs` and get rid of `person.rs` file.

FintanH: Sounds fine to me :+1:

xphoniex:
looks like the only place where `lastCommit` is used, is for `Blob`:
```
$ grep -rn "\<lastCommit\>" src/
src/project.ts:51:  lastCommit: CommitHeader;
src/project.ts:69:  lastCommit: CommitHeader;
src/project.ts:77:  lastCommit: CommitHeader;
src/project.ts:87:  lastCommit: CommitHeader;
src/base/projects/Blob.svelte:19:  const lastCommit = blob.lastCommit;
src/base/projects/Blob.svelte:203:        <div class="last-commit" title={lastCommit.author.name} use:twemoji>
src/base/projects/Blob.svelte:204:          <span class="hash">{lastCommit.sha1.slice(0, 7)}</span>
src/base/projects/Blob.svelte:205:          {lastCommit.summary}
```

radicle-surf/src/source/object/tree.rs
```diff
@@ -92,29 +73,50 @@ impl Serialize for Tree {
...
-    pub entry: directory::Entry,
+    name: String,
+    entry: Entry,
 }
 
-impl From<directory::Entry> for TreeEntry {
-    fn from(entry: directory::Entry) -> Self {
-        Self { entry }
+impl TreeEntry {
+    pub fn get_commit<C: ToCommit, P: AsRef<Path>>(
```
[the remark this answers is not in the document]
keepsimple1: It is a method of `TreeEntry`.  The idea is that, while iterating the tree's entries,  if a caller is interested in the `commit` (i.e. `lastCommit`) of the entry, they can call this method to obtain it. 

We could also remove this from public API and always retrieve `commit` in advance and embed it in the `TreeEntry` itself. I'm ok with either way conceptually and am in the process of trying them to see which one might work better in practice.

keepsimple1:
@FintanH
>  But I'm already being scolded https://github.com/radicle-dev/radicle-git/pull/83 for changing outputs of serialization 🙃 Unless it's easy to change on the frontend, we might need to stick with providing the path too.

Understood. One thing I wanted to bring up is that:  `httpd` should not return JSON solely based on the serialization result of `radicle-surf` API returns.  They should have a well defined JSON output for themselves and retrieve the info needed from `radicle-surf` API. 

i.e. `radicle-surf` API returns serialization => processed / filtered + other local info => `httpd` API JSON output. 

For example, I think the full path of an entry can (and should be) `prefix_path + entry_name` , where `prefix_path` is the input of `httpd` API, e.g. `tree_handler()`.  This also avoid a potential consistency between the input `prefix_path` and the `path` inside `Tree` if we were to store the path inside the `Tree`.   (Yes, they should always be same, but again, it's software ;-).

git-ref-format/core/src/deriv.rs
```diff
@@ -96,7 +96,7 @@ impl<'a> Qualified<'a> {
     /// Add a namespace.
     ///
     /// Creates a new [`Namespaced`] by prefxing `self` with
-    /// "refs/namespaces/<ns>".
+    /// "refs/namespaces/`ns`".
```
keepsimple1, line 99: due to CI error.

radicle-git-types/src/reference.rs
```diff
@@ -390,7 +390,7 @@ impl<N, R> Reference<N, R, Many> {
     where
         Self: ToString,
     {
-        ext::References::from_globs(repo, &[self.to_string()])
+        ext::References::from_globs(repo, [self.to_string()])
```
keepsimple1, line 393: due to CI error.

radicle-surf/src/source/object/blob.rs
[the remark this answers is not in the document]
keepsimple1: `name` and `path` are removed because they are not part of `Blob`, where `Blob` is an object of content. There could be multiple file names (paths) pointing to the same blob.

radicle-surf/src/source/object/tree.rs
```diff
@@ -92,29 +69,71 @@ impl Serialize for Tree {
     {
         const FIELDS: usize = 4;
         let mut state = serializer.serialize_struct("Tree", FIELDS)?;
+        state.serialize_field("oid", &self.id)?;
         state.serialize_field("entries", &self.entries)?;
         state.serialize_field("lastCommit", &self.commit)?;
-        state.serialize_field("name", &self.directory.name())?;
-        state.serialize_field("path", &self.directory.location())?;
         state.end()
```
keepsimple1, line 75: Similar to `Blob`, more than one directories can point to the same `Tree`.  Hence `Tree` (in git) itself does not own `name` or `path`.

radicle-surf/src/source/object/tree.rs
keepsimple1: I have included `lastCommit` now.

radicle-git-types/src/reference.rs
FintanH: I'm fixing these in a separate PR https://github.com/radicle-dev/radicle-git/pull/84

Better practice to separate exclusive changes into separate commits

radicle-surf/src/file_system/directory.rs
```diff
@@ -355,7 +358,8 @@ impl Directory {
 
         // Walks only the first level of entries.
         tree.walk(git2::TreeWalkMode::PreOrder, |path, entry| {
-            match Entry::from_entry(entry, Path::new(path).to_path_buf()) {
+            let prefix = self.path().join(path); // Must start at the parent path.
```
FintanH, line 361: I was under the impression that `path` always ends up being `""`, so I think this `join` doesn't do anything.

I explored that here https://github.com/radicle-dev/radicle-git/pull/83

radicle-surf/src/file_system/directory.rs
```diff
@@ -423,6 +427,8 @@ impl Directory {
     }
 
     /// Find the `Directory` found at `path`, if it exists.
+    ///
+    /// If `path` is `ROOT_DIR` (i.e. an empty path), returns self.
```
FintanH, line 431: :+1:

radicle-surf/src/git/repo.rs
```diff
@@ -252,9 +260,78 @@ impl Repository {
...
+
+    /// Returns a [`Tree`] for `path` in `commit`.
+    pub fn tree<C: ToCommit, P: AsRef<Path>>(&self, commit: C, path: &P) -> Result<Tree, Error> {
+        let commit = commit
+            .to_commit(self)
+            .map_err(|e| Error::ToCommit(e.into()))?;
+        let dir = self
+            .root_dir(commit.id)?
+            .find_directory(path, self)?
+            .ok_or_else(|| Error::PathNotFound(path.as_ref().to_path_buf()))?;
```
FintanH, line 289:
```suggestion
        let dir = self.directory(commit.id, path)?;
```

radicle-surf/src/git/repo.rs
```diff
@@ -252,9 +260,78 @@ impl Repository {
...
+            .map(|en| {
+                let name = en.name().to_string();
+                let path = en.path();
+                let commit = self
+                    .last_commit(&path, commit.id)?
+                    .ok_or(Error::PathNotFound(path))?;
+                let commit_header = Header::from(commit);
+                Ok(TreeEntry::new(name, en.into(), commit_header))
+            })
+            .collect();
```
FintanH, line 302:
```suggestion
        let entries = dir
            .entries(self)?
            .map(|en| {
                let name = en.name().to_string();
                let path = en.path();
                let commit = self
                    .last_commit(&path, commit.id)?
                    .ok_or(Error::PathNotFound(path))?;
                let commit_header = Header::from(commit);
                Ok(TreeEntry::new(name, en.into(), commit_header))
            })
            .collect::<Result<Vec<_>, _>>();
```

radicle-surf/src/git/repo.rs
```diff
@@ -252,9 +260,78 @@ impl Repository {
...
+
+    /// Returns a [`Blob`] for `path` in `commit`.
+    pub fn blob<C: ToCommit, P: AsRef<Path>>(&self, commit: C, path: &P) -> Result<Blob, Error> {
+        let commit = commit
+            .to_commit(self)
+            .map_err(|e| Error::ToCommit(e.into()))?;
+        let file = self
+            .root_dir(commit.id)?
+            .find_file(path, self)?
+            .ok_or_else(|| Error::PathNotFound(path.as_ref().to_path_buf()))?;
```
FintanH, line 321:
```suggestion
        let file = self.file(commit.id, path)?;
```

radicle-surf/src/source/object/tree.rs
```diff
@@ -126,16 +146,23 @@ impl Serialize for TreeEntry {
...
+                Entry::Tree(_) => "tree",
             },
         )?;
+        state.serialize_field(
+            "oid",
+            match self.entry {
+                Entry::Blob(id) => id.to_string(),
+                Entry::Tree(id) => id.to_string(),
+            }
+            .as_str(),
```
FintanH, line 163: I'd suggest adding `fn object_id` to `TreeEntry`, and `ext::Oid` already implements `Serialize` 
```suggestion
            self.object_id()
```

radicle-surf/src/source/object/blob.rs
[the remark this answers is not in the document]
FintanH: But a `path` is needed for the front end code, because they will be looking at the particular `Blob` for a given path. I don't think this is good enough, either we try to provide all the serialized data that's required or we provide none of it and leave it up to the `httpd` code to define the data they need and the serialization, but they compute that data using `radicle-surf`. This approach is only providing a halfway solution that leaves no one happy.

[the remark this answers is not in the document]
keepsimple1: > either we try to provide all the serialized data that's required or we provide none of it and leave it up to the httpd code to define the data they need and the serialization, but they compute that data using radicle-surf

I think it should be the latter. The fact that `radicle-surf` types support serialization does not mean `httpd` should rely on this serialization for their JSON output.

I will update this to avoid confusions.

radicle-git-types/src/reference.rs
keepsimple1: Thanks for fixing these. I've rebased.

radicle-surf/src/source/object/blob.rs
```diff
@@ -117,6 +91,24 @@ pub enum BlobContent {
...
+    /// Returns the size of this `BlobContent`.
+    pub fn size(&self) -> usize {
+        match self {
+            Self::Plain(content) => content.len(),
+            Self::Binary(bytes) => bytes.len(),
+        }
+    }
+
+    /// Returns the content as bytes.
+    pub fn as_bytes(&self) -> &[u8] {
```
keepsimple1, line 104: This is to support the idea that `httpd` (or other users) can / should serialize the content in any way they wanted.

radicle-surf/src/file_system/directory.rs
keepsimple1: > I was under the impression that path always ends up being "", so I think this join doesn't do anything.

Yes, this change's main purpose is to get `self.path()`, not the `path`, just like you explored. 

In general, the `path` might be "" or not "".  It is always "" in this case because we walk only  the first level, and skip any `Directory(_)` entry. 

I'm ok to go with either changes. If we go with your changes, I would propose to add a comment explaining why `path` is always "" and hence ignored.

radicle-surf/src/file_system/directory.rs
```diff
@@ -285,16 +289,21 @@ pub struct Directory {
     prefix: PathBuf,
     /// The object identifier of the git tree of this directory.
     id: Oid,
+    /// A commit that has this directory in its snapshot. It is used to find the
+    /// history of this directory.
+    head_commit: Oid,
```
FintanH, line 294: Is it always the head commit? Can it be the last seen commit for this directory? Can it be any old commit for this directory? "Head" has quite specific semantics, which is why I'm asking.

radicle-surf/src/file_system/directory.rs
```diff
@@ -71,6 +71,7 @@ pub mod error {
 /// The [`FileContent`] of a file can be retrieved via
 /// [`File::content`].
 #[derive(Clone, PartialEq, Eq, Debug)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
```
FintanH, line 74: Ok, I'm confused about our policy on serialization now. Are we leaving it up to applications to define it or are we defining it?

Btw, we never provide deserialization, I'm not sure we need to :)

radicle-surf/src/file_system/directory.rs
FintanH: I don't think we're ever going to end up with `path` not being empty, so I think it makes more sense to ignore it because it has no significance to us. 

I would say keep the change I made once you rebase and you can leave a comment if you'd like :)

radicle-surf/src/file_system/directory.rs
```diff
@@ -381,6 +396,15 @@ impl Directory {
         }
     }
 
+    /// Returns the last commit that created or modified this directory.
+    pub fn last_commit(&self, repo: &Repository) -> Result<Commit, error::Directory> {
+        // FIXME: it's messy to convert between error::Directory and repo::Error.
```
FintanH, line 401: This messiness tells me that this function is in the wrong place :)

radicle-surf/src/file_system/directory.rs
```diff
@@ -402,7 +426,9 @@ impl Directory {
             .ok_or_else(|| error::Directory::InvalidPath(path.to_string_lossy().to_string()))?;
 
         Ok(entry
-            .and_then(|entry| Entry::from_entry(&entry, parent.to_path_buf()).transpose())
+            .and_then(|entry| {
+                Entry::from_entry(&entry, parent.to_path_buf(), self.head_commit).transpose()
```
FintanH, line 430: I'm trying to convince myself that passing the `self.head_commit` makes sense here. I think a directory's contents can be changed but not necessarily all of its children.

For example, if you look [here](https://github.com/radicle-dev/radicle-git/tree/main/git-ref-format)  the commit doesn't affect all of that tree
image

radicle-surf/src/source/object/blob.rs
FintanH: I think we could just remove `BlobContent` and stick to a vector then :)

radicle-surf/src/source/object/blob.rs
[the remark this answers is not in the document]
FintanH: Ok, that makes sense to me but I believe this should be communicated to the rest of the clients team :) It might make sense to have a crate that goes into `radicle-git` for this so that the interface can be worked upon quickly -- although it then becomes separated from the `http` code.

Any thoughts @radicle-dev/clients?

radicle-surf/src/source/object/blob.rs
keepsimple1: I opened https://github.com/radicle-dev/radicle-git/issues/86 to track this.

radicle-surf/src/file_system/directory.rs
keepsimple1: `head_commit` seems not an ideal name. It is the last seen commit or any newer commit that contains this directory version with the particular `directory.id`.

With such a commit, we can easily find the history of this particular directory version, .e.g. finding out which commit created/updated this directory with its `oid` (a.k.a the last_commit).
 
I didn't want to store the `last_commit` directly (as in `Tree`) because `Directory::entries` is a lazy function, I wanted to have a "lazy" `last_commit()` as well.

We can try a different name, how about `later_commit`, or `repo_commit` ?

radicle-surf/src/file_system/directory.rs
keepsimple1: Are you referring to the implementation, or are you saying `Directory` should not have `last_commit()` method?

radicle-surf/src/file_system/directory.rs
keepsimple1: Sorry I didn't get it. `head_commit` does not mean changing all of the children.

radicle-surf/src/file_system/directory.rs
FintanH: I'm saying that this method should not exist in `Directory`, aye. 

> I think we need to refactor the error handling for radicle-surf. I've encountered a couple of times when errors from different modules would link to each other and cause circles.

When does this occur? That to me means that those methods should likely be in `Repository` because that's at the highest-level.

radicle-surf/src/file_system/directory.rs
FintanH: What I'm saying is that you're using the `head_commit` of the parent tree to populate the `head_commit` of all its children. As I mentioned above, I don't really get your intention of what `head_commit` is for, but if it's for displaying the last commit to touch an entry (be it another tree or file) then this is wrong.

keepsimple1: As discussed, I've updated to diff to include the `last_commit` directly in the struct of `Directory` and `File`, to avoid confusions. It will also be easier to retrieve them when needed.

radicle-surf/src/file_system/directory.rs
keepsimple1: I've updated the diff to store the last commit in the struct during construction.

radicle-surf/src/git/repo.rs
```diff
@@ -249,12 +257,72 @@ impl Repository {
...
+            .map(|en| {
+                let name = en.name().to_string();
+                let path = en.path();
+                let commit = self
+                    .last_commit(&path, commit.id)?
+                    .ok_or(Error::PathNotFound(path))?;
+                let commit_header = Header::from(commit);
+                Ok(TreeEntry::new(name, en.into(), commit_header))
+            })
+            .collect::<Result<Vec<TreeEntry>, Error>>()?;
```
FintanH, line 298: I don't think the `TreeEntry` or `Error` are necessary, not a blocker.
```suggestion
            .collect::<Result<Vec<_>, _>>()?;
```

radicle-surf/src/source/object/tree.rs
```diff
@@ -18,125 +18,215 @@
...
     where
         S: Serializer,
     {
         const FIELDS: usize = 4;
         let mut state = serializer.serialize_struct("Tree", FIELDS)?;
+        state.serialize_field("oid", &self.id)?;
         state.serialize_field("entries", &self.entries)?;
         state.serialize_field("lastCommit", &self.commit)?;
-        state.serialize_field("name", &self.directory.name())?;
-        state.serialize_field("path", &self.directory.path())?;
```
FintanH, line 99: Similarly, this is going to break compatibility.

radicle-surf/src/source/object/tree.rs
```diff
@@ -18,125 +18,215 @@
...
+    ///     "oid": "6d6240123a8d8ea8a8376610168a0a4bcb96afd0"
+    ///   },
+    /// ```
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
     where
         S: Serializer,
     {
         const FIELDS: usize = 4;
         let mut state = serializer.serialize_struct("TreeEntry", FIELDS)?;
-        state.serialize_field("path", &self.entry.path())?;
```
FintanH, line 130: And again

radicle-surf/src/git/repo.rs
keepsimple1: It didn't work for me.  I would get the following errors: 
```
error[E0282]: type annotations needed
   --> radicle-surf/src/git/repo.rs:296:17
    |
296 |                 Ok(TreeEntry::new(name, en.into(), commit_header))
    |                 ^^ cannot infer type of the type parameter `E` declared on the enum `Result`
    |
help: consider specifying the generic arguments
    |
296 |                 Ok::<object::tree::TreeEntry, E>(TreeEntry::new(name, en.into(), commit_header))
    |                   ++++++++++++++++++++++++++++++

error[E0283]: type annotations needed
   --> radicle-surf/src/git/repo.rs:296:17
    |
296 |                 Ok(TreeEntry::new(name, en.into(), commit_header))
    |                 ^^ cannot infer type of the type parameter `E` declared on the enum `Result`
297 |             })
298 |             .collect::<Result<Vec<_>, _>>()?;
    |                                            - type must be known at this point
```

radicle-surf/src/source/object/blob.rs
```diff
@@ -102,8 +78,6 @@ impl Serialize for Blob {
         state.serialize_field("binary", &self.is_binary())?;
         state.serialize_field("content", &self.content)?;
         state.serialize_field("lastCommit", &self.commit)?;
-        state.serialize_field("name", &self.file.name())?;
-        state.serialize_field("path", &self.file.path())?;
```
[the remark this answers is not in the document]
keepsimple1: Since `Blob` is only contents and does not have `name` or `path`, I propose that we land this change. And then I submit a PR for `radicle-httpd` to use our API return values instead of its serialization to construct their JSON output. That will also give a concrete example for solving issue https://github.com/radicle-dev/heartwood/issues/172. 

Currently `radicle-httpd` PR https://github.com/radicle-dev/heartwood/pull/165 plans to move to commit 79a94721366490053e2d8ac1c1afa14fb0c25f09.  So this diff won't break their code yet.

radicle-surf/src/git/repo.rs
FintanH: stoopid rust

PR #81, opened 2022-12-08
git-ref-format: fix to deserialization

FintanH:
When deserializing through `serde_json::from_value` the following
error would occur:

```
  panicked at 'called Result::unwrap() on an Err value: Error("invalid
  type: string "master", expected a borrowed string", line: 0, column:
  0)'
```

To fix this it is necessary for the Deserialize implementations to
deserialize into an owned String and convert from there.

While attempting to test this the `Qualified` and `Namespaced`
implementations were not able to be used:

```
  error: implementation of `serde::de::Deserialize` is not general enough
  --> git-ref-format/t/src/properties/name.rs:86:9
   |
86 |         roundtrip::json_value(qualified);
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ implementation of `serde::de::Deserialize` is not general enough
   |
   = note: `Qualified<'_>` must implement `serde::de::Deserialize<'0>`, for any lifetime `'0`...
   = note: ...but `Qualified<'_>` actually implements
   `serde::de::Deserialize<'1>`, for some specific lifetime `'1`
```

The Deserialize for both of those types now return owned data by
returning a 'static lifetime instead.

### DIFF
--- a/git-ref-format/core/src/serde.rs
+++ b/git-ref-format/core/src/serde.rs
@@ -42,7 +42,8 @@ impl<'de> Deserialize<'de> for RefString {
     where
         D: Deserializer<'de>,
     {
-        Deserialize::deserialize(deserializer).map(|x: &RefStr| x.to_owned())
+        Deserialize::deserialize(deserializer)
+            .and_then(|x: String| Self::try_from(x).map_err(de::Error::custom))
     }
 }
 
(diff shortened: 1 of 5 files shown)


cloudhead (approved):
LGTM

PR #88, opened 2022-12-21
surf: stats for a given revision

FintanH:
The `stats` method always provided the stats at the `head` of the repository.

Instead, allow the revision to be optionally specified and default to head if not provided.

### DIFF
--- a/radicle-surf/src/git/repo.rs
+++ b/radicle-surf/src/git/repo.rs
@@ -336,10 +336,20 @@ impl Repository {
         rev.to_commit(self)
     }
 
-    /// Gets the [`Stats`] of this repository.
+    /// Gets the [`Stats`] of this repository starting from the
+    /// `HEAD` (see [`Repository::head`]) of the repository.
     pub fn stats(&self) -> Result<Stats, Error> {
+        self.stats_from(&self.head()?)
+    }
+
+    /// Gets the [`Stats`] of this repository starting from the given
+    /// `rev`.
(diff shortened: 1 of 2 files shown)


xphoniex:
I'd personally prefer if we keep `repo.stats()` and add another fn, e.g. `repo.stats_for_head(sha1)`

FintanH:
> I'd personally prefer if we keep `repo.stats()` and add another fn, e.g. `repo.stats_for_head(sha1)`

Can you tell me why? I feel like it's unnecessary, but perhaps there's a good reason to.

xphoniex:
> Can you tell me why? I feel like it's unnecessary, but perhaps there's a good reason to.

I like to keep using `repo.stats()` and since you've added an optional rev as argument, now in order to get the stats of repo, one has to write `repo.stats(None)`, which both looks weird and doesn't add anything to context.  

Imagine reading this code in a few months and wondering what the `None` is supposed to be in this context. (also this is not backwards compatible)

Instead you can create a `pub fn stats_for_head<R>(&self, rev: &R)` and re-use the logic in `fn stats`:

```
pub fn stats<R>(&self, rev: Option<&R>) -> Result<Stats, Error>
    where
        R: Revision,
    {
        self.stats_for_head(self.head())
    }
```
✓ backwards compatible
✓ adds some context, as to what the arg is

FintanH:
Cheers, the arg argument makes sense. At the moment, I'm not so interested in compatibility for the API functions as we're still not at a release version yet.

xphoniex (approved):

PR #90, opened 2022-12-29
surf: fix comparison of tree entries

FintanH:
Comparing tree entries would only take into account their names. While this is useful, it is also useful to compare by their entry kind first, where trees take precedence over blobs.

### DIFF
--- a/radicle-surf/src/object/tree.rs
+++ b/radicle-surf/src/object/tree.rs
@@ -36,14 +36,16 @@ use crate::{fs, object::commit};
 pub struct Tree {
     /// The object id of this tree.
     id: Oid,
+    /// The first descendant entries for this tree.
     entries: Vec<TreeEntry>,
     /// The commit object that created this tree object.
     commit: commit::Header,
 }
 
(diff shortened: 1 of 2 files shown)


xphoniex (approved):

radicle-surf/src/object/tree.rs
```diff
@@ -106,12 +106,29 @@ impl Serialize for Tree {
...
     Blob(Oid),
 }
 
+impl PartialOrd for Entry {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for Entry {
```
keepsimple1, line 121: It will be useful to add a comment here or at the enum definition to describe the ordering for `Entry`.

keepsimple1 (approved):
LGTM. Could you please also add / update a test case?

xphoniex (approved):

keepsimple1 (approved):
Thanks for the updates! Looks good.

PR #92, opened 2023-01-04
surf: remove object::Error

FintanH:
The Error type is no longer used thanks to the clean up of the Tree and Blob constructions now being exposed through Repository.

Remove the Error type and clean up the imports as a result.


keepsimple1 (approved):
LGTM! Thanks.

Issue #91, opened 2023-01-04
radicle-surf: error handling refactoring


keepsimple1:
Currently, `radicle-surf` defines separate error types for each module (sometimes even [multiple error types in inside a module](https://github.com/radicle-dev/radicle-git/blob/9be2c3d1400ffd059a2e2c24fcc87938affbcdd7/radicle-surf/src/fs.rs#L36-L62)). 

The upside is that each error tells where (which module or type) it is from. 

The downside is that the error chaining could lead to circular chains, for example: 

[`object::Error`](https://github.com/radicle-dev/radicle-git/blob/9be2c3d1400ffd059a2e2c24fcc87938affbcdd7/radicle-surf/src/object.rs#L35-L49) can chain / wrap transparently `fs::error::Directory` and `git::Error`. Now `fs::error::Directory` or `git::Error` cannot (not using too much tricks) chain `object::Error`, otherwise it will cause a circle. The compile would fail.

Another downside is that , IMO, exposing many (more than one) error types from a single crate would add unnecessary complexity to its API and usage.

I'd like to propose to define a single public error type for `radicle-surf` crate that uses detailed error names to tell where (which module or type) the error is from.  The current [`git::Error`](https://github.com/radicle-dev/radicle-git/blob/9be2c3d1400ffd059a2e2c24fcc87938affbcdd7/radicle-surf/src/git/repo.rs#L58) can form the basis of such a type.

As an example, the current error `fs::error::Directory::InvalidPath(String)` can be moved into `git::Error::DirectoryInvalidPath(String)`.  i.e. the new error name has `Directory` as its prefix to indicate the module.

Further, if we get rid of `git/` and promote its contents, this public error type would just be `crate::Error`.

FintanH:
> [`object::Error`](https://github.com/radicle-dev/radicle-git/blob/9be2c3d1400ffd059a2e2c24fcc87938affbcdd7/radicle-surf/src/object.rs#L35-L49) can chain / wrap transparently `fs::error::Directory` and `git::Error`. Now `fs::error::Directory` or `git::Error` cannot (not using too much tricks) chain `object::Error`, otherwise it will cause a circle. The compile would fail.

Good news :) https://github.com/radicle-dev/radicle-git/pull/92

[...]

PR #93, opened 2023-01-05
remove last_commit() in Directory and File

keepsimple1:
It is not clear at this point that `Directory` and `File` needs a `last_commit()` method as the user likely has the context to call `repo.last_commit()` to achieve the goal. As it is easier to add things than to deleting things in future, we deemed it is better not to have them at this point.

Also renamed "id" to "tree_id" in Directory and "blob_id" in File so that their meanings are clearer.

This is to address one problem / FIXME mentioned in https://github.com/radicle-dev/radicle-git/issues/91#issuecomment-1370801069 . The diff also clarifies more between `Tree`/`Blob` and `Directory`/`File`.

### DIFF
--- a/radicle-surf/src/fs.rs
+++ b/radicle-surf/src/fs.rs
@@ -31,9 +31,11 @@ use git2::Blob;
 use radicle_git_ext::{is_not_found_err, Oid};
 use radicle_std_ext::result::ResultExt as _;
 
-use crate::{Commit, Repository, Revision};
+use crate::{Repository, Revision};
 
 pub mod error {
+    use std::path::PathBuf;
+
     use thiserror::Error;
 
     #[derive(Debug, Error, PartialEq)]
@@ -45,7 +47,7 @@ pub mod error {
         #[error(transparent)]
         File(#[from] File),
         #[error("the path {0} is not valid")]
-        InvalidPath(String),
+        InvalidPath(PathBuf),
     }
 
     #[derive(Debug, Error, PartialEq, Eq)]
@@ -58,6 +60,8 @@ pub mod error {
     pub enum File {
         #[error(transparent)]
         Git(#[from] git2::Error),
+        #[error("the path {0} is not valid")]
+        InvalidPath(PathBuf),
     }
 }
 
@@ -78,9 +82,7 @@ pub struct File {
     /// in respect to the root of the git repository.
     prefix: PathBuf,
     /// The object identifier of the git blob of this file.
-    id: Oid,
-    /// The commit that created this file version.
-    last_commit: Commit,
+    blob_id: Oid,
 }
 
 impl File {
@@ -90,7 +92,7 @@ impl File {
     /// so should not end in `name`.
     ///
     /// The `id` must point to a git blob.
-    pub(crate) fn new(name: String, prefix: PathBuf, id: Oid, last_commit: Commit) -> Self {
+    pub(crate) fn new(name: String, prefix: PathBuf, id: Oid) -> Self {
         debug_assert!(
             !prefix.ends_with(&name),
             "prefix = {:?}, name = {}",
@@ -100,8 +102,7 @@ impl File {
         Self {
             name,
             prefix,
-            id,
-            last_commit,
+            blob_id: id,
         }
     }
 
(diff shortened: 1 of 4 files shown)


radicle-surf/src/fs.rs
```diff
@@ -78,9 +82,9 @@ pub struct File {
     /// in respect to the root of the git repository.
     prefix: PathBuf,
     /// The object identifier of the git blob of this file.
-    id: Oid,
-    /// The commit that created this file version.
-    last_commit: Commit,
+    blob_id: Oid,
+    /// The commit id of a snapshot that has this file.
+    snapshot_id: Oid,
```
keepsimple1, line 87: I was debating if `parent_commit` is a better name, but eventually opted for `snapshot_id`. Let me know if any better names come up.

radicle-surf/src/fs.rs
```diff
@@ -136,9 +140,17 @@ impl File {
...
-        let blob = repo.find_blob(self.id)?;
+        let blob = repo.find_blob(self.blob_id)?;
         Ok(FileContent { blob })
     }
+
+    /// Returns the last commit that created or modified this file.
+    pub fn last_commit(&self, repo: &Repository) -> Result<Commit, error::File> {
+        let path = self.prefix.join(escaped_name(&self.name));
+        repo.last_commit(&path, self.snapshot_id)
+            .map_err(|_| error::File::InvalidPath(path.clone()))
```
keepsimple1, line 151: We are losing / ignoring the `git::Error` from `Repository::last_commit` due to the challenge of chaining it into `error::File`. 

We could remove this method in this module so we won't call `repo.last_commit()`, but the underlying general error chaining issue still exists.

radicle-surf/src/fs.rs
```diff
@@ -143,14 +148,6 @@ impl File {
         let blob = repo.find_blob(self.blob_id)?;
         Ok(FileContent { blob })
     }
-
-    /// Returns the last commit that created or modified this file.
-    pub fn last_commit(&self, repo: &Repository) -> Result<Commit, error::File> {
```
keepsimple1: It is possible to remove this method, but we still need to have `snapshot_id` so that we can call `repo.last_commit` later.

radicle-surf/t/src/file_system.rs
```diff
@@ -120,7 +120,10 @@ mod directory {
         let repo = Repository::open(GIT_PLATINUM).unwrap();
         let root = repo.root_dir(Branch::local(refname!("dev"))).unwrap();
         let dir = root.find_directory(&"this/is", &repo).unwrap().unwrap();
-        let last_commit = dir.last_commit(&repo).unwrap();
+        let last_commit = repo
+            .last_commit(&dir.path(), dir.snapshot_id())
+            .unwrap()
+            .unwrap();
```
keepsimple1, line 126: I think it's less ergonomic to call `repo.last_commit` compared with calling `dir.last_commit`.

FintanH (changes requested):
I'm not convinced `File` and `Directory` need `snapshot_id` as part of their structs. They can both be associated with many `Oid`s and the need for which is the last one will depend on the context of what the user is browsing -- generally some branch name or a certain commit they are looking at.

keepsimple1:
`snapshot_id` is a helper in the structs, in case the caller lost their context. Also, for a given `Directory`, calling `repo.last_commit()` with any context will return a fixed commit or None, because the underlying tree object is the same. 

That said, as we discussed, it is easier to add this `snapshot_id` or `last_commit()` method to `Directory` and `File` when they are needed. For now, we will not have them.

radicle-surf/src/fs.rs
```diff
@@ -58,6 +60,8 @@ pub mod error {
     pub enum File {
         #[error(transparent)]
         Git(#[from] git2::Error),
+        #[error("the path {0} is not valid")]
+        InvalidPath(PathBuf),
```
FintanH, line 64: This isn't used :)

radicle-surf/src/fs.rs
```diff
@@ -78,9 +82,7 @@ pub struct File {
     /// in respect to the root of the git repository.
     prefix: PathBuf,
     /// The object identifier of the git blob of this file.
-    id: Oid,
-    /// The commit that created this file version.
-    last_commit: Commit,
+    blob_id: Oid,
```
FintanH, line 85: I'd rather we just stick with the `id` naming

FintanH (approved):
I have some comments that I'm going to apply in the merge commit and push the changes :) Thanks for the work and conversation on these changes!

Issue #80, opened 2022-12-07
radicle-surf: flatten the src/ sub-directories?


keepsimple1:
It seems to me that the sub-directories in `radicle-surf/src` are not helping much.

```
├── src
│   ├── diff
│   │   └── git.rs
│   ├── diff.rs
│   ├── file_system
│   │   └── directory.rs
│   ├── file_system.rs
│   ├── git
│   │   ├── branch.rs
│   │   ├── commit.rs
│   │   ├── glob.rs
│   │   ├── history.rs
│   │   ├── namespace.rs
│   │   ├── repo
│   │   │   └── iter.rs
│   │   ├── repo.rs
│   │   ├── stats.rs
│   │   └── tag.rs
│   ├── git.rs
│   ├── lib.rs
│   ├── source
│   │   ├── commit.rs
│   │   ├── object
│   │   │   ├── blob.rs
│   │   │   └── tree.rs
│   │   ├── object.rs
│   │   └── person.rs
│   └── source.rs
````
For example: `file_system/directory.rs` can be fold into `file_system.rs`. 

`git/` directory is not really necessary as everything is based on git anyway. 

`source/`:  I mentioned before, I think it's hard to understand what exactly `source` means here and how it is different from other modules.

`diff/git.rs`: although it's okay to separate this part into its own file, it is not clear from this directory & filename why they are separated from `diff.rs`.

FintanH:
Ya, I think you're right -- I had a similar thought.

> For example: file_system/directory.rs can be fold into file_system.rs.

I think `directory` could move into `lib.rs`. We don't seem to need `file_system` really.

> git/ directory is not really necessary as everything is based on git anyway.
>
> source/: I mentioned before, I think it's hard to understand what exactly source means here and how it is different from other modules.

These two are interesting. I think you're right that it's strange to have a `git` since everything is now `git` based anyway. However, it does seem useful to differentiate it from `source`. Or we completely merge the two, but that requires some upfront work of building a more opinionated browsing experience, for example, a `Directory` (or `Tree`) carries extra metadata such as what was the last commit that touched it.

> diff/git.rs: although it's okay to separate this part into its own file, it is not clear from this directory & filename why they are separated from diff.rs.

The separation was due to multiple VCSs being potentially used. I do like having that extra file defining all the `git2` conversions though.

keepsimple1:
[...]

PR #98, opened 2023-01-17
move up blob.rs and tree.rs to src/

keepsimple1:
This is to solve issue #80


radicle-surf/src/tree.rs
```diff
@@ -109,50 +109,47 @@ impl Serialize for Tree {
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub enum Entry {
+pub enum TreeEntryKind {
```
FintanH, line 112: nit: What's the reason for the rename?

If I was to do renaming, I'd prefer this be `EntryKind` and `TreeEntry` as `Entry`. I would make the module `pub mod tree` that way if it needs to be qualified then people can use `tree::Entry` and `tree::EntryKind`.

keepsimple1: Yeah, the reason was for qualifying the name. I can do what you suggested ;-)

FintanH (approved):
Nice, thanks for making the changes :ok_hand:

Issue #86, opened 2022-12-19
radicle-surf: remove `BlobContent`


keepsimple1:
I think we could just remove `BlobContent` and stick to a vector then :)

_Originally posted by @FintanH in https://github.com/radicle-dev/radicle-git/pull/79#discussion_r1052112024_

keepsimple1:
This is done.

PR #99, opened 2023-01-19
change BlobContent to borrow its content from git2::Blob

keepsimple1:
This patch is triggered by https://github.com/radicle-dev/radicle-git/issues/86, but later morphed into a different one. I think it's okay to keep a type `BlobContent`, but I wanted to avoid always copying the content bytes that could be potentially quite large.


FintanH:
My worry here is that `BlobContent` _needs_ to live as long as the `Repository` that created it. But maybe that's a fair requirement.

radicle-surf/src/object/blob.rs
```diff
@@ -83,53 +81,34 @@ impl Serialize for Blob {
...
 
 /// Variants of blob content.
-#[derive(PartialEq, Eq)]
-pub enum BlobContent {
-    /// Content is plain text and can be passed as a string.
-    Plain(String),
-    /// Content is binary and needs special treatment.
-    Binary(Vec<u8>),
+pub struct BlobContent<'a> {
+    git2_blob: git2::Blob<'a>,
```
FintanH, line 85: We tend to name these kinds of things `inner`:
```suggestion
    inner: git2::Blob<'a>,
```

keepsimple1:
> My worry here is that `BlobContent` _needs_ to live as long as the `Repository` that created it. But maybe that's a fair requirement.

I also have this concern on the back of my mind. Maybe this is an optimization that we don't need yet. I'll think about it a bit more.

keepsimple1:
I have added a `BlobVec` to support `Blob::to_owned()`.  Now our repo API would return a `Blob<BlobRef<'a>>` first, and if the user needed, they can call `.to_owned()` to get an owned content bytes. I hope this can address our concern. Let me know what you think.

keepsimple1:
I've rebased this PR to the latest `main`.

keepsimple1:
please let me know if any blockers for landing this. @FintanH

PR #102, opened 2023-01-27
radicle-surf: refactoring the error type

keepsimple1:
This is to solve issue #91. 

Moved `Error` in `repo.rs` out into a crate level `Error` in `error.rs`. Repository specific errors are wrapped by the `crate::Error` as well.

Also, the latest Rust 1.67.0 causes many lint errors with `println!`. I am changing `lint` to stay with Rust `1.66.1` to avoid unnecessary lint errors. We can move to Rust 1.67.0 or `stable` via a later PR.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -21,7 +21,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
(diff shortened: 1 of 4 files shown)


FintanH (approved):
I'm happy with the compromise :+1:

Issue #40, opened 2022-10-26
radicle-surf: support options in diff function


keepsimple1:
- Support diff between two revisions for a given file. One way to do it is to support different options in our diff function(s), in addition to `from` and `to` revisions.

- Can we support binary flag check that is currently disabled?

FintanH:
> * Can we support binary flag check that is [currently disabled](https://github.com/radicle-dev/radicle-git/blob/ea0afc49935906376860d65c60faa8c11e278869/radicle-surf/src/vcs/git/repo.rs#L424)?

I'm not entirely sure if we _want_ to support it. A diff between binary files is not really meaningful to a human :smile:

keepsimple1:
> * Support diff between two revisions for a given file. One way to do it is to support different options in our [diff function(s)](https://github.com/radicle-dev/radicle-git/blob/ea0afc49935906376860d65c60faa8c11e278869/radicle-surf/src/vcs/git/repo.rs#L158), in addition to `from` and `to` revisions.

Is there a real use case for getting a diff between two revisions for a given file?  If yes, I think one option is to support filtering a `Diff` by a path.  Does that make sense?

FintanH:
> > * Support diff between two revisions for a given file. One way to do it is to support different options in our [diff function(s)](https://github.com/radicle-dev/radicle-git/blob/ea0afc49935906376860d65c60faa8c11e278869/radicle-surf/src/vcs/git/repo.rs#L158), in addition to `from` and `to` revisions.
> 
> Is there a real use case for getting a diff between two revisions for a given file? If yes, I think one option is to support filtering a `Diff` by a path. Does that make sense?

I dunno if there's any existing use case but I would say that it could be deemed useful. I, personally, like checking the diff between two revisions on a single file from time to time. 

> If yes, I think one option is to support filtering a Diff by a path. Does that make sense?

I think that could work, something like `Diff::find(&self, path: &Path) -> [FileDiff](https://github.com/radicle-dev/radicle-git/blob/main/radicle-surf/src/diff.rs#L150-L163)`. I'd be interested in seeing how it looks defined on `Repository` as well, something like `diff_file(&self, to: &R, from: &R, path: &Path) -> FileDiff`. It might be more performant using the underlying `git2` functionality.

`FileDiff` won't quite work though, because of not knowing what kind of changes were made, i.e. was it added, deleted, modified, copied, or moved? 

[...]

PR #106, opened 2023-01-30
add diff_file method for Repository

keepsimple1:
This is to solve issue #40 .

This patch turns out larger than I expected as I ended up to refactor the `FileDiff` type.  The serialization of `Diff` is kept compatible.

A new branch `diff-test` for `git-platinum` was created for testing diff serialization.


radicle-surf/src/diff.rs
```diff
@@ -34,11 +35,11 @@ pub mod git;
...
-    pub added: Vec<Added>,
-    pub deleted: Vec<Deleted>,
-    pub moved: Vec<Moved>,
-    pub copied: Vec<Copied>,
-    pub modified: Vec<Modified>,
+    pub added: Vec<FileDiff>,
+    pub deleted: Vec<FileDiff>,
+    pub moved: Vec<FileDiff>,
+    pub copied: Vec<FileDiff>,
+    pub modified: Vec<FileDiff>,
```
FintanH, line 42: There are two things I'd like to explore further in this PR :)
1) I think the separation of fields becomes useless in this representation. The reason I say this is that theoretically `added` `FileDiff`s could be of `DiffKind::Deleted`. I'm thinking that we could have a single field called `files` and we could add methods `added`, `deleted`, etc. that gives a filtered iterator of those specific diffs.
2) Continuing on from 1) I think we could have `fields` be of type `BTreeMap<PathBuf, FileDiff>`, which a) enforces that each `PathBuf` has only one `FileDiff` and b) the lookup is easy -- just call `get`.

keepsimple1: For 1, it makes sense to me. Updated the diff. 

For 2, it turns out that `DiffKind::Moved` has two `PathBuf`s: `old_path` and `new_path`. If we call `diff_file`, we can find a `FileDiff` for both of them, i.e. one `Added` for `new_path`, and one `Deleted` for `old_path`.  So it is difficult to pick one as the key in `BTreeMap` for such `FileDiff`.  Given that we don't really do queries at this moment, I will leave this change for  future investigations.

radicle-surf/src/diff.rs
```diff
@@ -154,12 +199,104 @@ pub struct Modified {
...
+impl FileDiff {
+    pub fn path_string(&self) -> String {
+        match &self.kind {
+            DiffKind::Added(x) => x.path.to_string_lossy().to_string(),
+            DiffKind::Deleted(x) => x.path.to_string_lossy().to_string(),
+            DiffKind::Modified(x) => x.path.to_string_lossy().to_string(),
+            DiffKind::Moved(x) => format!("{:?} -> {:?}", &x.old_path, &x.new_path),
+            DiffKind::Copied(x) => format!("{:?} -> {:?}", &x.old_path, &x.new_path),
+        }
+    }
```
FintanH, line 238: This is only used for testing, I'm not convinced it's that useful.

I'd rather we implemented a nice `Display` instead. Preferably compatible with Git's diff output.

FintanH (commented):
I think this is fine for now. There are a few things that bug me at the moment but we can figure it out in a follow-up. I'll list them here:
1) The diff contents and kinds can be mismatched, e.g. if I match on the copied kind and contents then I have to ignore binary or plain because they wouldn't make any sense.
2) The `added`, `deleted`, etc. methods return FileDiff which if matched on have variants that need to be ignored.
3) I know `EofNewLine` is existing code, but I feel like it can happen with added files too. Something to look into.

radicle-surf/src/diff.rs
FintanH: I see, and ya it has that nice effect of not needing the full directory/file name.

keepsimple1:
> 1. The diff contents and kinds can be mismatched, e.g. if I match on the copied kind and contents then I have to ignore binary or plain because they wouldn't make any sense.

As Git detects `moved` / `copied` based on a similarity threshold that can be < 100%, the `DiffContent` might not always be empty for the copied kind.

I did a test locally with git, and saw that happening:  
(readme3.md is a new file copied off and modified a bit, then I used the threshold of 90% for copy detection). 
```
$ diff emoji.txt readme3.md
2a3
> a line.

$ git log --summary -C90 -1
commit f2c4e04263ea182f7a811bfc9a82addab62870e5 (HEAD -> diff-test)
Author: Han Xu <keepsimple@gmail.com>
Date:   Tue Jan 31 05:14:20 2023 -0800

    readme3.md

 copy emoji.txt => readme3.md (94%)
```

keepsimple1:
> 2\. The `added`, `deleted`, etc. methods return FileDiff which if matched on have variants that need to be ignored.

Fixed this one. Now they return the types `Added`, `Deleted`, etc.

keepsimple1:
> 3\. I know `EofNewLine` is existing code, but I feel like it can happen with added files too. Something to look into.

Agreed.

radicle-surf/src/diff.rs
keepsimple1: I've removed `path_string()` as it is no longer needed.

FintanH:
> > 1. The diff contents and kinds can be mismatched, e.g. if I match on the copied kind and contents then I have to ignore binary or plain because they wouldn't make any sense.
> 
> As Git detects `moved` / `copied` based on a similarity threshold that can be < 100%, the `DiffContent` might not always be empty for the copied kind.
> 
> I did a test locally with git, and saw that happening: (readme3.md is a new file copied off and modified a bit, then I used the threshold of 90% for copy detection).
> 
> ```
> $ diff emoji.txt readme3.md
> 2a3
> > a line.
> 
> $ git log --summary -C90 -1
> commit f2c4e04263ea182f7a811bfc9a82addab62870e5 (HEAD -> diff-test)
> Author: Han Xu <keepsimple@gmail.com>
> Date:   Tue Jan 31 05:14:20 2023 -0800
> 
>     readme3.md
> 
>  copy emoji.txt => readme3.md (94%)
> ```
> 
> (Btw, I had to enable copy detection via `git config --global --add diff.renames copy`)

Great point, and nice find :)

FintanH (approved):
Nice, looks good :+1:

PR #107, opened 2023-01-31
surf: borrowed and owned Blob content

FintanH:
Avoid copying the bytes content of a `git2::Blob` by introducing a generic parameter on Blob.

The parameter can either be a `BlobRef` or `BlobVec`, by using the `to_owned` function.

`BlobRef` and `BlobVec` both implement `AsRef` and `Deref` for `[u8]` to make them easier to use and allows the definition of generic functions on `T`.

### DIFF
--- a/radicle-surf/src/blob.rs
+++ b/radicle-surf/src/blob.rs
@@ -18,6 +18,8 @@
 //! Represents git object type 'blob', i.e. actual file contents.
 //! See git [doc](https://git-scm.com/book/en/v2/Git-Internals-Git-Objects) for more details.
 
+use std::ops::Deref;
+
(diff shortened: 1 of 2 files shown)


FintanH:
Improvements on top of https://github.com/radicle-dev/radicle-git/pull/99 :) thanks for laying the ground work :+1:

keepsimple1:
> Improvements on top of #99 :) thanks for laying the ground work 👍

I think it's better for me to land #99 first:  it will provide clearer authorship and history, as well as be easier to review changes.  You can always submit / update improvements after that.

keepsimple1 (approved):
Nice!

Issue #108, opened 2023-02-06
radicle-surf: how does `EofNewLine` work?


keepsimple1:
> 3\. I know `EofNewLine` is existing code, but I feel like it can happen with added files too. Something to look into.

`EofNewLine` is a bit confusing to me. I'd like to investigate this in a separate PR.

_Originally posted by @keepsimple1 in https://github.com/radicle-dev/radicle-git/issues/106#issuecomment-1414477715_

PR #109, opened 2023-02-06
radicle-surf: refactoring EofNewLine

keepsimple1:
This is to solve issue #108 .

- Moved `eof` into `DiffContent::Plain`  as it only makes sense to plain text diffs. This changes the serialization of `Diff` and `DiffContent` but it's limited to only `eof` element.
- Added a variant in `EofNewLine` enum: `NoneMissing` that is used when both the old file and the new file have a EOF new line, or for a newly `Added` file etc.

### DIFF
--- a/radicle-surf/src/diff.rs
+++ b/radicle-surf/src/diff.rs
@@ -90,16 +90,8 @@ impl Diff {
         &self.stats
     }
 
-    fn insert_modified(
-        &mut self,
-        path: PathBuf,
-        hunks: impl Into<Hunks<Modification>>,
-        eof: Option<EofNewLine>,
-    ) {
-        let diff = DiffContent::Plain {
-            hunks: hunks.into(),
-        };
-        let diff = FileDiff::Modified(Modified { path, eof, diff });
+    fn insert_modified(&mut self, path: PathBuf, diff: DiffContent) {
+        let diff = FileDiff::Modified(Modified { path, diff });
         self.files.push(diff)
     }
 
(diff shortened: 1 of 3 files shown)


radicle-surf/src/diff.rs
```diff
@@ -90,16 +90,11 @@ impl Diff {
...
-    ) {
-        let diff = DiffContent::Plain {
-            hunks: hunks.into(),
-        };
-        let diff = FileDiff::Modified(Modified { path, eof, diff });
+    fn insert_modified(&mut self, path: PathBuf, diff_content: DiffContent) {
+        let diff = FileDiff::Modified(Modified {
+            path,
+            diff: diff_content,
+        });
```
FintanH, line 97: small nit pick:
```suggestion
    fn insert_modified(&mut self, path: PathBuf, diff: DiffContent) {
        let diff = FileDiff::Modified(Modified {
            path,
            diff,
        });
```

FintanH (approved):
Looks good other than a small nit :+1:

cloudhead:
cc @xphoniex any changes needed on httpd?

xphoniex:
> cc @xphoniex any changes needed on httpd?

eof tests live here in `surf`, and they seem to be returning what is expected of them, also [updated the methods](https://github.com/radicle-dev/radicle-git/pull/109/commits/0e224da4b31273fc9581fadf26938d7c326fb29c) for the disabled ones.

---

as for updating `surf` for `httpd`, needs more work, currently working on it.

Issue #100, opened 2023-01-24
radicle-surf:  towards v0.9


keepsimple1:
Although there are some open items (like error handling), I think we can start to work towards the first new version (v0.9) of this crate, including fixing up the doc, stabilizing the methods in API. (The current version on crate.io is [v0.8](https://crates.io/crates/radicle-surf/0.8.0)). 

(Regarding the new version number, I think v0.9 is better than v1.0 as the API might evolves a bit based on feedbacks. Normally v1.0 means the API is deemed to be stable for a long while).

Tasks before publishing v0.9: 

- [x] Add CHANGELOG file.
- [x] Review public APIs including their `cargo doc`.
- [x] Review dependencies: including the dependencies from `radicle-git` repo, such dependencies should be published already.
    - [x] `radicle-std-ext`
    - [x] `radicle-git-ext`
    - [x] `git-ref-format`

keepsimple1:
v0.9 is published 👍

PR #111, opened 2023-02-07
radicle-surf: update docs and add test for commit signature

keepsimple1:
Work related to #100 .

- Clean up some doc comments and add a missing test case.
- Use `crates.io` for specifying the internal crate dependencies. (Removing `path`)

### DIFF
--- a/radicle-surf/Cargo.toml
+++ b/radicle-surf/Cargo.toml
@@ -49,7 +49,6 @@ features = ["serde"]
 
 [dependencies.radicle-std-ext]
 version = "0.1.0"
-path = "../radicle-std-ext"
 
 [dependencies.serde]
 version = "1"
--- a/radicle-surf/src/lib.rs
+++ b/radicle-surf/src/lib.rs
@@ -36,7 +36,7 @@
 pub use git_ref_format;
 
 /// Represents an object id in Git. Re-exported from `radicle-git-ext`.
-pub use radicle_git_ext::Oid;
+pub type Oid = radicle_git_ext::Oid;
 
 pub mod blob;
 pub mod diff;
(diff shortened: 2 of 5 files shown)


radicle-surf/src/lib.rs
```diff
@@ -36,7 +36,7 @@
 pub use git_ref_format;
 
 /// Represents an object id in Git. Re-exported from `radicle-git-ext`.
-pub use radicle_git_ext::Oid;
+pub type Oid = radicle_git_ext::Oid;
```
keepsimple1, line 39: This is to avoid `cargo doc` copying the doc comment from the original imported type directly.

radicle-surf/Cargo.toml
```diff
@@ -49,7 +49,6 @@ features = ["serde"]
 
 [dependencies.radicle-std-ext]
 version = "0.1.0"
-path = "../radicle-std-ext"
```
FintanH, line 52: Why is this being removed? :thinking:

keepsimple1: In my understanding, having both `path` and `version` introduces ambiguity: what happens if `path` is not compatible with `version`? It makes it possible that the published version 0.9.0 `radicle-surf` uses a different `radicle-std-ext` from the local version 0.9.0.

FintanH: I'm pretty sure that's _not_ how it works. It allows for local development of `radicle-std-ext` and the actual version on crates.io is checked when running `cargo publish`.

You can see more information here https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html.

Everything else looks good, so I'll merge this and amend the merge commit putting the `path` back in.

PR #115, opened 2023-02-09
git-commit: remove useless check in to_string()

keepsimple1:
This diff is triggered by a test failure in PR https://github.com/radicle-dev/heartwood/pull/280 .

~~However, it is not clear to me what exactly format we want for a commit. I'm adding some comments in the code.~~

This diff morphs into a simple refactoring for a check that I think is useless.

### DIFF
--- a/git-commit/src/lib.rs
+++ b/git-commit/src/lib.rs
@@ -292,12 +292,8 @@ impl ToString for Commit {
         if !self.trailers.is_empty() {
             writeln!(buf).ok();
         }
-        for (i, trailer) in self.trailers.iter().enumerate() {
-            if i < self.trailers.len() {
-                writeln!(buf, "{}", Trailer::from(trailer).display(": ")).ok();
-            } else {
-                write!(buf, "{}", Trailer::from(trailer).display(": ")).ok();
-            }
+        for trailer in self.trailers.iter() {
+            writeln!(buf, "{}", Trailer::from(trailer).display(": ")).ok();
         }
         buf
     }


git-commit/src/lib.rs
```diff
@@ -289,9 +289,6 @@ impl ToString for Commit {
         write!(buf, "{}", self.message.trim()).ok();
         writeln!(buf).ok();
 
-        if !self.trailers.is_empty() {
-            writeln!(buf).ok();
-        }
         for (i, trailer) in self.trailers.iter().enumerate() {
             if i < self.trailers.len() {
                 writeln!(buf, "{}", Trailer::from(trailer).display(": ")).ok();
```
keepsimple1, line 294: We are doing different things for the last trailer vs not-last trailers. I am curious what is the reason?

git-commit/src/lib.rs
```diff
@@ -289,9 +289,6 @@ impl ToString for Commit {
         write!(buf, "{}", self.message.trim()).ok();
         writeln!(buf).ok();
 
-        if !self.trailers.is_empty() {
-            writeln!(buf).ok();
-        }
```
keepsimple1, line 294: For the extra newline, I think it's simpler to remove this one instead of the one above.

keepsimple1:
After looking at failed tests and the `to_string()` output, I think this diff is probably not necessary. But I think it's worth to discuss what we want to do going forward.

git-commit/src/lib.rs
keepsimple1: no longer valid.

git-commit/src/lib.rs
```diff
@@ -292,12 +292,8 @@ impl ToString for Commit {
         if !self.trailers.is_empty() {
             writeln!(buf).ok();
         }
-        for (i, trailer) in self.trailers.iter().enumerate() {
-            if i < self.trailers.len() {
```
keepsimple1, line 296: This will always be true, right?

FintanH: Ya, you're right. I must have left it in during a refactoring :sweat_smile: It seems like we always want the new line as well, so this looks good :+1:

FintanH (approved):
Good catch :ok_hand:

Issue #114, opened 2023-02-09
Merging git crates


FintanH:
It was suggested in Discord that merging the Git crates would ease development and publishing:

> cloudhead
 —  
you don't have to publish 3 crates, only 1, and you don't have to update the versions between all three crates to match
and there's only one version number to track
so I know what state the library is in, vs. having to check three different revisions
not having to publish 3 crates also means you don't have to worry about the order in which you publish the crates..
the only situation in which none of this matters is if you aren't publishing to crates.io
then you just have 1 rev to track
>
> fintohaps
 — 
Today at 2:19 PM
Oh, so you're suggesting having one big crate with all the other crates as modules? That doesn't seem great if, for example, radicle-cob relies on git-ref-format but then also has to pull in radicle-surf
>
> cloudhead
 — 
Today at 2:35 PM
yeah, probably we can merge some of them at least
eg. git-commit + git-trailers + git-ref-format
something like that
and keep surf separate
oh and radicle-git-ext
there's just way too many
>
> fintohaps
 — 
Today at 3:18 PM
Sure, can look into it

One thing to note here is that surf only uses a subset of these -- doesn't use `git-trailers` or `git-commit`. This begs the question, should it depend on the single crates or the mega crate?

keepsimple1:
> One thing to note here is that surf only uses a subset of these -- doesn't use `git-trailers` or `git-commit`. This begs the question, should it depend on the single crates or the mega crate?

I suggest we merge `git-trailers` and `git-commit` into the `radicle-surf` crate for two reasons: 

1. Currently `radicle-surf` does not support trailers in the commit messages. This will bring in a new feature I think it's valuable.
2. `git-commit` crate overlaps with `Commit` type in `radicle-surf`. We can combine them and get rid of the overlapping.

cloudhead:
From the perspective of heartwood:

* radicle-httpd depends on `surf` only
* radicle library depends on `git-ref-format`, `radicle-git-ext`
* radicle-cob depends on `git-trailers`, `git-ref-format`, `git-commit`, `radicle-git-ext`
* radicle-node depends on `git-ref-format`

`git-ref-format` seems to be the most widely used.
`git-trailers` I'm not sure why it even exists, we parse trailers in ~20 lines of code in side `radicle`, using only `git2`, and I would like to get rid of the giant macro in `radicle-cob` which uses it.
[...]

PR #116, opened 2023-02-14
git-commit: refactor trailers: remove git-trailers depedency.

keepsimple1:
This patch is to take a stab at one area in issue #114 :  simplify or remove git-trailers as a dependency crate.

Includes a new method `trailers_of_key()` that can potentially replace [this file](https://github.com/radicle-dev/heartwood/blob/master/radicle-cob/src/trailers.rs).  (See test case)

### DIFF
--- a/git-commit/Cargo.toml
+++ b/git-commit/Cargo.toml
@@ -17,7 +17,3 @@ thiserror = "1"
 version = "0.16.1"
 default-features = false
 features = ["vendored-libgit2"]
-
-[dependencies.git-trailers]
-version = "0.1.0"
-path = "../git-trailers"
--- a/git-commit/src/lib.rs
+++ b/git-commit/src/lib.rs
@@ -17,13 +17,18 @@ use std::{
 };
 
 use git2::{ObjectType, Oid};
-use git_trailers::{self as trailers, OwnedTrailer, Trailer};
 
 pub mod author;
 pub use author::Author;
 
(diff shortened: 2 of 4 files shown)


FintanH (changes requested):
I appreciate you exploring this, but in my opinion, I don't see it as an improvement over using git-trailers.

I find the declarative version, via `nom`, much better compared to an iterative loop with lots of mutating states.

The original point made by @cloudhead was:
> git-trailers I'm not sure why it even exists, we parse trailers in ~20 lines of code in side radicle, using only git2, and I would like to get rid of the giant macro in radicle-cob which uses it.

The code in `radicle` that's being referred to is [here](https://github.com/radicle-dev/heartwood/blob/master/radicle/src/storage/git.rs#L633-L652), which uses [`git2::message_trailers_strs`](https://docs.rs/git2/0.16.1/git2/fn.message_trailers_strs.html).

If we're going to get rid of the `git-trailers` code then I think the `git2` function should be used.

keepsimple1:
> If we're going to get rid of the `git-trailers` code then I think the `git2` function should be used.

I looked at git2 functions but did not use them because they kept the message as is, i.e. not separating them.  I'm OK with using git2 function if we keep the original message, or separate the message later.

keepsimple1:
> Are you happy to close this and we improve on #117 if needed?

Yes, no problem. Closing this one now.

PR #117, opened 2023-02-14
git-commit: remove dependency on git-trailers

FintanH:
Supersedes #116, as it uses `git2`'s functionality for parsing commit trailers.

The git-trailers crate depends on parsing logic using nom. This is a heavy dependency for something that can be done using git2.

Remove the dependency of git-trailers in git-commit and localise the trailer types -- using git2 to parse the trailers out of the git commit message.

### DIFF
--- a/git-commit/src/lib.rs
+++ b/git-commit/src/lib.rs
@@ -17,13 +17,15 @@ use std::{
 };
 
 use git2::{ObjectType, Oid};
-use git_trailers::{self as trailers, OwnedTrailer, Trailer};
 
 pub mod author;
 pub use author::Author;
 
 pub mod headers;
 pub use headers::{Headers, Signature};
+
+pub mod trailers;
+pub use trailers::{OwnedTrailer, Trailer, Trailers};
 
 /// A git commit in its object description form, i.e. the output of
 /// `git cat-file` for a commit object.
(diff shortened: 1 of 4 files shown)


cloudhead:
Looking good. Should we include a test?

git-commit/src/lib.rs
```diff
@@ -243,27 +243,22 @@ impl FromStr for Commit {
...
-                    (message, trailers)
-                },
-            },
-        );
+        let trailers = trailers::Trailers::parse(message).map_err(error::Parse::Trailers)?;
+
+        let message = message
+            .strip_suffix(&trailers.to_string(": "))
+            .unwrap_or(message)
+            .to_string();
```
keepsimple1, line 251: Git messages could have a divider line (`---`) followed by a patch part. Will `heartwood` support such messages?

keepsimple1 (commented):
Overall looks good, with an inline comment. Are the existing test cases sufficient for this change?

git-commit/src/lib.rs
FintanH: So, I believe the idea is that `git-interpret-trailers` works on both commits and `.patch` files generated by `git format-patch`. The latter can also contain trailers and then the patch information is below the divider line, i.e. `---`.

That leads me to believe that we won't and shouldn't actually see `---` in git commit messages, and if we do then it was intentional by the operator and we shouldn't attempt to parse it.

FintanH:
Regarding the questions about tests, I believe [these](https://github.com/radicle-dev/radicle-git/blob/main/git-commit/t/src/commit.rs#L5-L45) should suffice. It covers the cases of no, one, and two trailers :)

git-commit/src/lib.rs
keepsimple1: Okay.

keepsimple1 (approved):

PR #118, opened 2023-02-17
radicle-git-ext: merge in git-commit

keepsimple1:
This is to address issue #114 .

- Merged `git-commit` into `radicle-git-ext` crate. Copied:

`author.rs`
`commit.rs`
`headers.rs`

`t/src/commit.rs`

`git-commit` crate is not deleted because it's a published crate. Some users might depend on it.

- Kept `git-ref-format` crate for now.

This is because `git-ref-format-macro`  uses `proc-macro` that AFAIK cannot mix into the same crate as `radicle-git-ext`. I don't have a good idea to work around this limit at this point. Maybe in future?

- Removed `reference.rs` in `racidle-git-ext` as it overlaps with `git-ref-format`.  Also removed `radicle-macros` and `radicle-git-types` crates that used types in `reference.rs`. 

Note that `radicle-macros` and `radicle-git-types` are not published on crates.io, hence I think it's safe to remove them. Let me know otherwise.


FintanH:
> This is because `git-ref-format-macro` uses `proc-macro` that AFAIK cannot mix into the same crate as `radicle-git-ext`. I don't have a good idea to work around this limit at this point. Maybe in future?

Can you explain what the issues were?

radicle-git-ext/src/lib.rs
```diff
@@ -5,18 +5,20 @@
 
 //! Extensions and wrappers for `git2` types
 
+pub mod author;
 pub mod blob;
+pub mod commit;
 pub mod error;
+pub mod headers;
```
FintanH, line 12: I would put `headers` as a submodule of `commit`.

radicle-git-ext/src/lib.rs
```diff
@@ -5,18 +5,20 @@
...
 pub mod oid;
-pub mod reference;
 pub mod revwalk;
+pub mod trailers;
```
FintanH, line 15: The same for `trailers`.

FintanH (commented):
Cool! Nice bit of clean up :) Couple of small nits to discuss before merging

keepsimple1:
> > This is because `git-ref-format-macro` uses `proc-macro` that AFAIK cannot mix into the same crate as `radicle-git-ext`. I don't have a good idea to work around this limit at this point. Maybe in future?
> 
> Can you explain what the issues were?

There are two issues I encountered: (I'm a newbie in proc-macros).

1.  proc-macros have to be defined in root source files in a crate. This is relatively easy to solve: we can put them under `src/` in this crate.
2. A crate cannot define both public proc-macros and public regular functions. 

After researching a bit more, I found that we can also solve item 2 by treating `git-ref-format-macro` like an internal crate, and re-export all public macros in the parent crate.  Such pattern is used in some common crates, like [in `thiserror`](https://github.com/dtolnay/thiserror/blob/a2d1ed1ccfc2a5dbb2a8fb45d4f938175a28bc86/src/lib.rs#L245).

keepsimple1:
I've tried out a patch for moving `git-ref-format` under `radicle-git-ext`, but it is quite involved. I think it's better to do that in a separate PR. I will post that patch in a new PR after this PR is landed.

FintanH (approved):
Sweet, thanks for that!

PR #119, opened 2023-02-21
move git-ref-format to under radicle-git-ext

keepsimple1:
This is one possible way to move `git-ref-format` crate to under `radicle-git-ext`, as proposed in issue #114 . The changes are:

- `git-ref-format` becomes an "internal" crate of `radicle-git-ext`, i.e. it is no longer intended to be used directly. Instead, should do this: `use radicle_git_ext::git_ref_format`. 
- `git-ref-format-macro` crate now has a few places importing from `radicle_git_ext::git_ref_format`. 
- All tests of `git-ref-format` are moved under `radicle-git-ext/t`.

Why did I keep `git-ref-format` as a (internal) crate instead of merging into `radicle-git-ext` completely? The reason is that `git-ref-format-macro` is a proc-macro crate that cannot easily merge into a regular crate. 

On the other hand, this means we would still need to publish more than one crate. So the benefit of this patch is debatable. Please let me know what you think.


radicle-git-ext/git-ref-format/macro/src/lib.rs
```diff
@@ -32,7 +32,7 @@ pub fn refname(input: TokenStream) -> TokenStream {
             let expand = quote! {
                 unsafe {
                     use ::std::mem::transmute;
-                    use ::git_ref_format::RefString;
+                    use radicle_git_ext::git_ref_format::RefString;
```
FintanH, line 35:
```suggestion
                    use ::radicle_git_ext::git_ref_format::RefString;
```

radicle-git-ext/src/lib.rs
```diff
@@ -20,3 +20,5 @@ pub use oid::*;
 pub use revwalk::*;
 pub use transport::*;
 pub use tree::Tree;
+
+pub use git_ref_format;
```
FintanH, line 24: Can we reexport this like:
```rust
pub use git_ref_format as ref_format
```
or
```rust
pub use git_ref_format as format
```

I find the `git_` part a bit verbose.

radicle-surf/src/refs.rs
```diff
@@ -6,7 +6,14 @@ use std::{
     convert::TryFrom as _,
 };
 
-use git_ref_format::{lit, name::Components, Component, Qualified, RefString};
+use radicle_git_ext::git_ref_format::{
```
FintanH, line 9: This is making me want to declare:
```rust
extern crate radicle_git_ext as git_ext
```
in `lib.rs` just so we can import as (also using the suggestion above):
```rust
use git_ext::ref_format::{ /* etc */ }
```

FintanH (commented):
Can we not make the proc-macro for `radicle-git-ext` instead? Or did you explore that route already?

keepsimple1:
> Can we not make the proc-macro for `radicle-git-ext` instead? Or did you explore that route already?

I tried but it didn't work. To demonstrate such errors, I added a test pub fn in `git-ref-format-macro` crate, and got this error message:

```
error: `proc-macro` crate types currently cannot export any items other than functions tagged with `#[proc_macro]`, `#[proc_macro_derive]`, or `#[proc_macro_attribute]`
   --> git-ref-format/macro/src/lib.rs:170:1
    |
170 | pub fn this_is_a_test() {
    | ^^^^^^^^^^^^^^^^^^^^^^^
```

radicle-surf/src/refs.rs
keepsimple1: I'm aliasing `git-ref-format` as `ref-format`, let me know if that will be enough, or you also want to alias `radicle-git-ext`.

PR #122, opened 2023-03-09
commit: verify data

FintanH:
The `Oid`s passed into `Commit::new` for parents and tree are never
verified to be commits or a tree, respectively.

Upon `write` of the commit, read the object using the ODB to determine
its type. If its the incorrect object then function fails without
writing to the ODB.


cloudhead (approved):
Nice.

keepsimple1 (approved):
Looks good!

PR #124, opened 2023-03-27
radicle-surf: Repository.blob_at() to retrieve a blob using its oid.

keepsimple1:
This is motived by a [discussion on zulip](https://radicle.zulipchat.com/#narrow/stream/369277-heartwood/topic/.E2.9C.94.20raw.20endpoint.20with.20oids).


radicle-surf/src/repo.rs
```diff
@@ -301,6 +304,20 @@ impl Repository {
         Ok(Blob::<BlobRef<'a>>::new(file.id(), git2_blob, last_commit))
     }
 
+    /// Retrieves the blob with `oid` in `commit`.
+    pub fn blob_at<'a, C: ToCommit>(
+        &'a self,
+        commit: C,
```
keepsimple1, line 310: The input `commit` is not required if we only wanted to retrieve the blob content.  It is here as we also need to retrieve the `Blob.commit` that created the blob (oid).

FintanH: I wasn't expecting to have the `commit` as a parameter, so what if we changed the result type to just `BlobRef<'a>`?

FintanH: > From the zulip discussion linked in Description, the caller would (most likely) have the commit info available.

My reading of the discussion was that we should be able to get a blob _without_ the commit[[0]]. So if something doesn't know about a commit but does have a blob id, then it can just retrieve the blob content.

[0]: https://radicle.zulipchat.com/#narrow/stream/369277-heartwood/topic/.E2.9C.94.20raw.20endpoint.20with.20oids/near/343707845

keepsimple1: I was referring to the original[ request message](https://radicle.zulipchat.com/#narrow/stream/369277-heartwood/topic/.E2.9C.94.20raw.20endpoint.20with.20oids/near/343705723). 

Yeah, like we discussed so far, there are two options that I can see:
 
a) Get the blob content only (input: `Oid`).  If the caller wanted to find the commit that created the blob, they make another HTTP call.  This is not consistent with existing `blob()` method.

b) Get the blob content and the commit that created the blob.  (input: `Oid` and a `Commit` that has the blob. ) .  Both info are retrieved in one call.  This is consistent with existing `blob()` method.

For the blob content part, b) does more involved work than a),  because it also retrieves the commit. 

It's a trade-off. My preference is to go with option b) based on the reasons I mentioned earlier. And if we did want to go with a), I would suggest to re-visit `Blob` struct and see if it makes sense to separate `Blob.commit` out, and then change `blob()` to be consistent with `blob_at()`.

keepsimple1: > I don't mind going down this route, but I don't understand why we have to go fetch the commit that created the blob. Can you explain why that's necessary?

It's because this existing struct and its API: https://github.com/radicle-dev/radicle-git/blob/6de53f72d4dda7fce78b2bec496159119ab186d3/radicle-surf/src/blob.rs#L52-L54

PR #129, opened 2023-05-24
Add file information to diff

cloudhead:
Adds file mode and file OIDs to the file diff.

### DIFF
--- a/radicle-surf/src/diff.rs
+++ b/radicle-surf/src/diff.rs
@@ -22,6 +22,8 @@ use std::{borrow::Cow, path::PathBuf, string::FromUtf8Error};
 #[cfg(feature = "serde")]
 use serde::{ser, ser::SerializeStruct, Serialize, Serializer};
 
+use git_ext::Oid;
+
 pub mod git;
 
 /// The serializable representation of a `git diff`.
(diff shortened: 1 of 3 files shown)


cloudhead:
Might need a `FileMode -> u32` too :thinking:

radicle-surf/src/diff.rs
```diff
@@ -229,6 +240,32 @@ impl DiffContent {
...
+    Blob,
+    /// For regular files that are executable.
+    BlobExecutable,
+    /// For directories.
+    Tree,
+    /// For symbolic links.
+    Link,
+    /// Used for Git submodules.
+    Commit,
+}
```
FintanH, line 257: Should we just wrap the `git2::FileMode`? Or is there a reason we don't?

FintanH (approved):
Generally looks good :+1:

radicle-surf/src/diff.rs
cloudhead: Good question. There are some extra modes in `git2::FileMode` that I don't know if we want to support because they are not actual Git file modes, those are `Unreadable` and `BlobGroupWritable`. The first one just means that the file is not readable, I think that is better handled somewhere else, if at all. And the latter is there for legacy reasons in git2.

PR #130, opened 2023-06-22
surf: Add hunk lines ranges

cloudhead:


radicle-surf/src/diff/git.rs
```diff
@@ -179,7 +179,12 @@ impl TryFrom<git2::Patch<'_>> for DiffContent {
                 let line = Modification::try_from(line)?;
                 lines.push(line);
             }
-            hunks.push(Hunk { header, lines });
+            hunks.push(Hunk {
+                header,
+                lines,
+                old: hunk.old_start()..hunk.old_start() + hunk.old_lines(),
```
FintanH, line 185: Just to clarify, will this cut off the last line in the range? :thinking: Or do you think it works counting the first line?

FintanH (commented):
Just a quick question, otherwise it looks good :+1:

radicle-surf/src/diff/git.rs
cloudhead: It shouldn't if I understand: say you've removed lines `4, 5`, then it'll be `4..6`, and since it's a non-inclusive range, it doesn't include line 6, which is correct.

Does that make sense?

PR #131, opened 2023-06-28
surf: Add copy detection

sebastinez:
This PR adds copy detection of files that were changed in the same Changeset.
To be able to test that change we also add a new copied file in the git-platinum fixture.


radicle-surf/src/repo.rs
```diff
@@ -519,6 +519,7 @@ impl Repository {
         // Detect renames by default.
         let mut find_opts = git2::DiffFindOptions::new();
         find_opts.renames(true);
+        find_opts.copies(true);
```
FintanH, line 522: I thought always finding copies was an expensive computation, so we were going to make it optional. Or did I misunderstand?

sebastinez: as far as I understand it, finding copies if the source of the copy operation has also been modified, isn't expensive.
See: https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---find-copiesltngt

The expensive operation is to look for copies when the source of the copy operation has been unmodified.
See: https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---find-copies-harder

One might argue that copy operations of modified files are quite rare, but I think since it's not that expensive we could still add it

FintanH: Ah ok, that makes sense :)

FintanH (dismissed):

PR #136, opened 2023-07-24
surf: Add `FileDiff::path` method

sebastinez:
This allows to get the latest path of a `FileDiff`, directly from `radicle_surf`

Usage example: In the `radicle-httpd` diff endpoints we want to add the blobs of the files that are involved in a diff, and this way we could do a `repo.blob(commit, file_diff.path())` to get the `path` quickly.


FintanH:
Looks good :+1: I'll merge it now :)

PR #141, opened 2023-08-30
Improved `Moved`

### LINKED ISSUE #140: `diff::Moved` might require object IDs and file modes
A recent [Radicle Patch](https://app.radicle.xyz/nodes/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/commits/c55badf0b8bee38e425d297decbd3560fad3c5e7) for heartwood indicates git files which are moved can have oids and file modes.

```
git show c55badf0b8bee38e425d297decbd3560fad3c5e7 | grep -A5 b/radicle/src/node/tracking/config.rs
diff --git a/radicle-node/src/service/tracking.rs b/radicle/src/node/tracking/config.rs
similarity index 96%
rename from radicle-node/src/service/tracking.rs
rename to radicle/src/node/tracking/config.rs
index cbc843c8..3f69208f 100644
```

However `diff::Moved` uses `PathBuf` as fields, where DiffFile might be better:

```
pub struct Moved {
    /// The old path to this file, relative to the repository root.
    pub old_path: PathBuf,
    /// The new path to this file, relative to the repository root.
    pub new_path: PathBuf,
    pub diff: DiffContent,
}
```

FintanH:
In some cases a moved file can also be modified and in these cases there will be a diff as well as new Oid's for the two sides of the diff.

Add DiffFile to the Moved type, changing the insert_moved method to accept the DiffFile's and the DiffContent.

When serializing a Moved, if the old and the new are the same then the serialization of the diff files and content are skipped -- same as the old behaviour. Otherwise, the two sides are serialized along with the content.

Fixed #140 

cc @sebastinez how will this affect radicle-interface/radicle-httpd for showing diffs for moved files?

### DIFF
--- a/radicle-surf/src/diff.rs
+++ b/radicle-surf/src/diff.rs
@@ -129,12 +129,21 @@ impl Diff {
         self.files.push(diff);
     }
 
-    pub fn insert_moved(&mut self, old_path: PathBuf, new_path: PathBuf) {
+    pub fn insert_moved(
+        &mut self,
+        old_path: PathBuf,
+        new_path: PathBuf,
+        old: DiffFile,
+        new: DiffFile,
+        content: DiffContent,
+    ) {
         self.update_stats(&DiffContent::Empty);
         let diff = FileDiff::Moved(Moved {
             old_path,
             new_path,
-            diff: DiffContent::Empty,
+            old,
+            new,
+            diff: content,
         });
         self.files.push(diff);
     }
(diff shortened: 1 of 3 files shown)


sebastinez:
Thanks @FintanH for the heads up, as far as I can see we would need to adapt the `radicle-interface` to display the diffs for moved files that have modifications.
But it wouldn't break `radicle-httpd` nor `radicle-interface`

sebastinez (approved):
Looks good to me, will add these changes to `radicle-interface`

FintanH:
> Thanks @FintanH for the heads up, as far as I can see we would need to adapt the `radicle-interface` to display the diffs for moved files that have modifications. But it wouldn't break `radicle-httpd` nor `radicle-interface`
> 
> I'll add a issue on `radicle-interface` to add this change to our types

That's great! Thanks for confirming that :)

PR #145, opened 2023-09-29
surf: Add `stats` property to each `DiffContent::Plain`

sebastinez:
This allows us to get the additions and deletions for each file in a diff.

### DIFF
--- a/radicle-surf/src/diff.rs
+++ b/radicle-surf/src/diff.rs
@@ -288,6 +288,7 @@ pub enum DiffContent {
     #[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
     Plain {
         hunks: Hunks<Modification>,
+        stats: FileStats,
         eof: EofNewLine,
     },
     Empty,
@@ -296,8 +297,16 @@ pub enum DiffContent {
 impl DiffContent {
     pub fn eof(&self) -> Option<EofNewLine> {
         match self {
-            Self::Plain { hunks: _, eof } => Some(eof.clone()),
+            Self::Plain { eof, .. } => Some(eof.clone()),
             _ => None,
+        }
+    }
+
+    pub fn stats(&self) -> Option<&FileStats> {
+        match &self {
+            DiffContent::Plain { stats, .. } => Some(stats),
+            DiffContent::Empty => None,
+            DiffContent::Binary => None,
         }
     }
 }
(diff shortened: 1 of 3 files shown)


radicle-surf/src/diff.rs
```diff
@@ -296,10 +297,39 @@ pub enum DiffContent {
 impl DiffContent {
     pub fn eof(&self) -> Option<EofNewLine> {
         match self {
-            Self::Plain { hunks: _, eof } => Some(eof.clone()),
+            Self::Plain {
+                hunks: _,
+                stats: _,
+                eof,
```
FintanH, line 303:
```suggestion
                eof, ..
```
I'd say we just do this :)

radicle-surf/src/diff.rs
```diff
@@ -296,10 +297,39 @@ pub enum DiffContent {
...
+            } => Some(eof.clone()),
             _ => None,
         }
     }
+    pub fn stats(&self) -> FileStats {
```
FintanH, line 308:
```suggestion

    pub fn stats(&self) -> FileStats {
```
New line please :pray:

radicle-surf/src/diff.rs
```diff
@@ -296,10 +297,39 @@ pub enum DiffContent {
...
+        let mut additions = 0;
+        let mut deletions = 0;
+
+        match &self {
```
FintanH, line 312: Wait, why are we calculating the stats here if it's contained in `DiffContent::Plain`? :thinking: 

~I think we want to see if `git2` can provide the stats per file and get them from there when we create the `DiffContent` (in the `TryFrom` impls under `diff::git`).~

So we should just match and return:
```rust
DiffContent::Plain { stats, .. } => Some(stats),
DiffContent::Empty => None,
DiffContent::Binary => None,
```
I'd return an `Option` because stats don't make sense for the other two.

radicle-surf/src/diff/git.rs
```diff
@@ -166,11 +169,19 @@ impl TryFrom<git2::Patch<'_>> for DiffContent {
                         old_missing_eof = true;
                         continue;
                     },
+                    git2::DiffLineType::Addition => {
```
FintanH, line 172: Ah you are calculating them here. I checked `DiffFile` too and there were no individual stats.

FintanH (changes requested):

radicle-surf/src/diff.rs
sebastinez: Yeah that works, seems like it did the job twice 😁

radicle-surf/src/diff/git.rs
sebastinez: I don't think we want the individual `stats` to show up in the `DiffFile` since we are interested in the changes introduced by a specific `diff` which is specified by `DiffContent`

radicle-surf/t/src/diff.rs
```diff
@@ -99,6 +104,10 @@ fn test_diff_file() -> Result<(), Error> {
                 new: 1..3,
             }]
             .into(),
+            stats: FileStats {
+                additions:2,
+                deletions:1
```
FintanH, line 109:
```suggestion
                additions: 2,
                deletions: 1
```

FintanH (approved):
Just a little fix and then happy to merge :+1:

radicle-surf/t/src/diff.rs
sebastinez: Thanks for the catch, is resolved 👍

